PR #7, opened 2023-05-05
AWS Module

squioc:
Add the AWS Module with its connectors


squioc:
The module is now in the repository

PR #11, opened 2023-05-31
Add CI workflow to GitHub actions

a-leonardi:
The action will perform the following tasks:
- lint code using ruff
- lint code using mypy
- lint code using black
- run pyton tests using pytest
- compute tests coverage using codecov
- upload test results to codecov

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,74 @@
+---
+name: CI
+on:
+  push:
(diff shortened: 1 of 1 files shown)


Darkheir (changes requested):
We already have a workflow doing a few things like running test, black and mypy.
The main missing part is codecov

Also the checks must be done by module and not for the full repository

PR #12, opened 2023-06-05
Trellix edr

vg-svitla:


Trellix/client/http_client.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        search_status_url = self._edr_searches_job_status_url(search_id)
+        while True:
+            headers = self._request_headers(scopes)
+            async with self.session().get(search_status_url, headers=headers) as status_response:
+                if status_response.status == 303:
+                    break
+
+                # In documentation it is said that we should wait 15 seconds before next request
+                await asyncio.sleep(15)
```
squioc, line 159: Could you limit the number of attempts to avoid infinite loop if the api isn't able to return a result?
The number of attempt could be pass as an argument, and you can replace the `while True:` by `for _ in range(max_attempts):`.

Trellix/trellix/connector.py
```diff
@@ -0,0 +1,73 @@
...
+            self.push_events_to_intakes([orjson.dumps(event).decode("utf-8") for event in events])
+
+        return events
+
+    def run(self) -> None:
+        """Runs TrellixEdr."""
+        loop = asyncio.get_event_loop()
+        try:
+            while self.running:
+                loop.run_until_complete(self.get_trellix_edr_events())
```
squioc, line 66: The connector only fetches edr_events? That about real-time searches and historical searches?

Trellix/client/http_client.py
```diff
@@ -0,0 +1,242 @@
...
+        query_url = self._edr_historical_searches_query_url
+        scopes = Scope.set_for_historical_searches()
+
+        headers = await self._request_headers(scopes)
+
+        # TODO check attributes once it have any data
+        query_body = {
+            "data": {
+                "type": "historicalSearches",
+                "attributes": {"startTime": "2023-06-03T09:26:22Z", "endTime": "2023-06-04T09:26:22Z"},
```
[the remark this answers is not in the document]
vg-svitla: Yes it should be. It was only for testing purposes.
Fixed

Trellix/client/http_client.py
vg-svitla: The problem with this solution is that we do not know how much time will take to process this task.
I mean the flow is the next -
1. Create job
2. Wait until job change status to success/fail. It means the response status code will be 303
And to be honest I do now know how much time it will take. I found better solution with 
`aiolimiter.AsyncLimiter` instead of `asyncio.sleep`
https://aiolimiter.readthedocs.io/en/latest/
And implemented it is Salesforce integration, so will use it here also.

But still it will not give us a solution of avoiding while True statement as for me :(

Trellix/trellix/connector.py
vg-svitla: Realtime searches less documented ( the same as historical searches ).
So I just tested that it returns empty response, provide api calls and did not include it yet into connector processing :)

vg-svitla: The main problem is that it might return the same data.
Historical search we should use only on first startup, after - realtime search might be preferable. 
But still was not able to collect responses to take a look if it contains the same data in terms of events

Trellix/trellix/connector.py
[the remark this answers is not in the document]
vg-svitla: Sure. Tests will populate. 
Did not provide tests yet for HttpClient also, as response data might be different and logic for handling it also might be different.

Trellix/client/http_client.py
```diff
@@ -0,0 +1,242 @@
...
+            data = await results_response.json()
+
+        return [TrellixEdrResponse[SearchRealtimeAttributes](**result) for result in data["data"]]
+
+    async def get_edr_investigations(self) -> List[TrellixEdrResponse[InvestigationAttributes]]:
+        """
+        All necessary workflow to get realtime searches.
+
+        Documentation for more details:
+            https://docs.trellix.com/fr/bundle/mvision-endpoint-detection-and-response-product-guide/page/GUID-D08D9CF3-45E5-4EB2-A155-A618730BDA46.html#GUID-D08D9CF3-45E5-4EB2-A155-A618730BDA46
```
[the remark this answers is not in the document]
vg-svitla: Thank you! Fixed

Trellix/pyproject.toml
```diff
@@ -0,0 +1,88 @@
...
+name = "SEKOIA.IO Automation Module to work with Trellix EDR events"
+version = "0.0"
+description = ""
+authors = []
+
+[tool.poetry.dependencies]
+python = ">=3.10,<3.12"
+orjson = "^3.7.7"
+python-dateutil = "^2.8.2"
+sekoia-automation-sdk = "^1.1.2"
```
squioc, line 11: I think we can upgrade `sekoia-automation-sdk' version number

squioc (changes requested):
Nice Job.

Some little clean up remaining and I think we are good.

squioc (approved):
LGTM

PR #66, opened 2023-06-13
Salesforce integration

vg-svitla:


Salesforce/client/token_refresher.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+        Refresh token based on class configuration.
+
+        Also triggers token refresh task.
+        """
+        params = {
+            "grant_type": "client_credentials",
+        }
+
+        url = URL("{0}/services/oauth2/token".format(self.auth_url)).with_query(urlencode(params, encoding="utf-8"))
```
squioc, line 111: Instead of the `format` method, you can also use [f-string](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings)
```suggestion
        url = URL(f"{self.auth_url}/services/oauth2/token").with_query(urlencode(params, encoding="utf-8"))
```

Salesforce/client/token_refresher.py
```diff
@@ -0,0 +1,160 @@
...
+        async with self.session().post(url, auth=BasicAuth(self.client_id, self.client_secret), json={}) as response:
+            response_data = await response.json()
+
+            self._token = SalesforceToken(
+                token=HttpToken(**response_data),
+                created_at=time.time(),
+                ttl=self.token_ttl,
+            )
+
+            await self._schedule_token_refresh(self._token.ttl)
```
squioc, line 122: How long last Salesforce's tokens?
We usually refresh the token 300 seconds (5 minutes) before it expires (e.g.: [CrowdStrike Auth](https://github.com/SEKOIA-IO/automation-library/blob/main/CrowdStrikeFalcon/crowdstrike_falcon/client/auth.py#L46)) to avoid negative responses due to authorization expiration.
```suggestion
            await self._schedule_token_refresh(self._token.ttl)
```

Salesforce/salesforce/connector.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        now = datetime.now(timezone.utc)
+        one_week_ago = now - timedelta(days=7)
+
+        with self.context as cache:
+            last_event_date_str = cache.get("last_event_date")
+
+            # if undefined, retrieve events from the last 7 days
+            if last_event_date_str is None:
+                return one_week_ago
```
squioc, line 70: I think that getting the events from the last hour, if no last_event_date exists, is preferable. I worry that customers, setting this integration, doesn't understand why they are not viewing their most recent events, in the interface, when starting the connector.

Salesforce/salesforce/connector.py
```diff
@@ -0,0 +1,178 @@
...
+        last_event_date = None
+        log_files = await self.salesforce_client.get_log_files(self.from_date.isoformat())
+        for log_file in log_files.records:
+            log_file_date = isoparse(log_file.LogDate)
+            if last_event_date is None or last_event_date < log_file_date:
+                last_event_date = log_file_date
+
+            records, csv_path = await self.salesforce_client.get_log_file_content(
+                log_file_uri=log_file.LogFile,
+                temp_dir=self.module.configuration.temp_files_dir,
```
squioc, line 144: I think that could be dangerous to allow customers to define the path to a temporary directory. It's preferable to use an existing directory like `/tmp`
```suggestion
```

squioc (changes requested):
Could you create the [manifest file](https://github.com/SEKOIA-IO/automation-library/blob/main/CrowdStrike_Telemetry/manifest.json) for the module and the [manifest file](https://github.com/SEKOIA-IO/automation-library/blob/main/CrowdStrike_Telemetry/trigger_crowdstrike_telemetry_events.json) for the connector?

squioc (commented):
Could you also add some [metrics](https://github.com/SEKOIA-IO/automation-library/blob/main/Google/google_module/metrics.py) in the connector?
Thanks

Salesforce/client/token_refresher.py
vg-svitla: I tried to find some time ago what is the best way of doing it in python, and found interesting discussion here
https://github.com/wemake-services/wemake-python-styleguide/issues/32
What do you think? 
I would like to make it more/less standard, so if we decide about f-strings, I'll use it everywhere

Salesforce/client/token_refresher.py
vg-svitla: Salesforce tokens have lifetime period of 2 hours. 
The problem is that they do not have any field in response that will identify expiration time, something like `expire_in`, `lifetime` or something similar.
Also there is an option to parse this token as JWT, it should have all info inside but I'm not sure if it is good idea
I'll setup refresh to 5 mins, but yes

Salesforce/salesforce/connector.py
vg-svitla: Yes make sense. Thank you!

Salesforce/salesforce/metrics.py
```diff
@@ -0,0 +1,25 @@
...
+# Declare prometheus metrics
+prom_namespace = "symphony_module_salesforce_events"
+
+OUTCOMING_EVENTS = Counter(
+    name="forwarded_events",
+    documentation="Number of events forwarded to SEKOIA.IO",
+    namespace=prom_namespace,
+    labelnames=["intake_key"],
+)
+
```
squioc, line 12:
```suggestion

# Declare common prometheus metrics
prom_namespace = "symphony_module_common"

```

Salesforce/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "properties": {},
+        "title": "Salesforce module Configuration",
+        "type": "object"
+    },
+    "description": "Salesforce provides users comprehensive tools to manage customer data, automate processes, analyze data and insights, and create personalized customer experiences. Salesforce also offers a variety of solutions for customer service, marketing automation, commerce, app development, and more",
+    "name": "Salesforce",
+    "uuid": "626c6110-0a7a-11ee-be56-0242ac120002",
+    "slug": "salesforce",
+    "version": "1.22"
```
squioc, line 12: We start module version with `1.0.0` :D
```suggestion
    "version": "1.0.0"
```

Salesforce/client/http_client.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Returns:
+            SalesforceEventLogFilesResponse:
+        """
+        logger.info("Getting log files from Salesforce. Start date is {0}", start_from)
+
+        query = self._log_files_query(start_from)
+
+        url = self._request_url_with_query(query)
+        headers = await self._request_headers()
```
squioc, line 204: Cannot we add the `Accept-Encoding: gzip` header to manage heavy responses? ([see](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_rest_compression.htm#intro_rest_compression))

squioc (changes requested):

Salesforce/client/token_refresher.py
vg-svitla: Yes. It is working solution. I've changed it to 5 mins, but token by itself is valid for 2 hours

squioc: 3600 was great. Just a mismatch between the comment and the implementation.

squioc (approved):

PR #94, opened 2023-06-22
fix for non json events

Adamowoc:
Fix for AWS non json incoming events 
> sekoia_automation.module - CRITICAL - unexpected character: line 1 column 1 (char 0)

### DIFF
--- a/AWS/aws/s3/queued.py
+++ b/AWS/aws/s3/queued.py
@@ -73,8 +73,7 @@ class AWSS3QueuedConnector(AWSConnector, metaclass=ABCMeta):
         raise NotImplementedError()
 
     def get_next_messages(self, **kwargs) -> list[dict]:
-        response = self.sqs_client.receive_message(
-            QueueUrl=self.queue_url, **kwargs)
+        response = self.sqs_client.receive_message(QueueUrl=self.queue_url, **kwargs)
         return response.get("Messages", [])
 
     def get_next_objects(self, messages: list) -> Generator[AWSS3Object, None, None]:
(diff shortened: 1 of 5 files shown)


PierrickV (commented):
Codecov reported an issue on the code that you added. I wonder if your test triggers the issue as excepted

AWS/aws/s3/queued.py
```diff
@@ -79,14 +79,18 @@ def get_next_messages(self, **kwargs) -> list[dict]:
...
+                for record in decoded_messages:
+                    try:
+                        yield AWSS3Object.from_record(self.s3_client, record)
+
+                        if receipt := message.get("ReceiptHandle"):
+                            receipts.append(receipt)
+                    except ValueError as e:
+                        self.log_exception(e, message="Invalid record")
+            except ValueError as e:
+                self.log_exception(e, message=f"Invalid JSON in message.\nInvalid message is: {message}")
```
gaelmuller, line 93: The except should probably be at line 84 since it is only covering the json deserialization (same for other instances)

gaelmuller (approved):

PierrickV (commented):
You still have the issue where the SQS code does not seem tested
image

PierrickV:
The error despaired when I reloaded :shrug: 

This other error is not tested btw https://app.codecov.io/gh/SEKOIA-IO/automation-library/pull/94/blob/AWS/aws/s3/queued.py#L94

PierrickV (approved):
Seems good to me

It would be nice to improve the coverage of this code :smile:

PR #95, opened 2023-06-22
Remove python-Levenshtein dependency

a-leonardi:
Reimplement Levenshtein distance so that we can remove the dependeyc from python-Levenshtein in CertificateTransparency This library's licence GPLv2 is not compatible with the licence under which CertificateTransparency is distributed


Darkheir (commented):
Code looks good

You should bump the version of the module in the `manifest.json` file

Darkheir (approved):

PR #116, opened 2023-07-13
[Cleanup] Salesforce

vg-svitla:
Cleanup
Change logic to work with CreatedDate instead of LogDate 
Add logs


squioc (changes requested):
Could you update the version number of the module?

squioc (approved):

PR #127, opened 2023-07-17
Logs also the query

TOUFIKIzakarya:
Hi @squioc, forgot to log also the query.


squioc (approved):
For testing purposes, I accept that log, but I think we should remove it before deploying the connector.

PR #130, opened 2023-07-17
Sophos: Change the way to handle exceptions in the EDR connector

squioc:
Change the way to handle exceptions in the EDR connector


TOUFIKIzakarya (approved):
LGFM, but I think you have to apply black.

PR #134, opened 2023-07-18
Add time log

TOUFIKIzakarya:
- Log the query time


squioc (changes requested):
Update the version number in manifest.json

squioc (approved):

PR #135, opened 2023-07-18
Split salesforce configuration

vg-svitla:


Salesforce/manifest.json
```diff
@@ -1,13 +1,35 @@
...
+                "type": "string"
+            }
+        },
+        "required": [
+            "client_id",
+            "client_secret",
+            "base_url"
+        ],
+        "secrets": [
+            "client_id",
```
squioc, line 24: I think we don't to define `client_id` as a secret.
```suggestion
```

squioc (approved):

PR #138, opened 2023-07-19
Fix Duo bugs

lvoloshyn-sekoia:


squioc (changes requested):
The code looks good to me.
Could you update the version number in the manifest and add an entry in CHANGELOG.md?
Thanks.

squioc (approved):

PR #147, opened 2023-07-26
Set default hostname for Vade Cloud

lvoloshyn-sekoia:


squioc (changes requested):
Could you update the version number in the manifest?

squioc (approved):

PR #148, opened 2023-07-26
Fix logging for Vade Cloud API authorization

lvoloshyn-sekoia:


VadeCloud/vadecloud_modules/client/auth.py
squioc, line 1: I wonder if specialized Exception are not preferable (see [Cybereason Authentication](https://github.com/SEKOIA-IO/automation-library/blob/main/Cybereason/cybereason_modules/client/auth.py)). I'm not a huge fan of that callback.

VadeCloud/manifest.json
```diff
@@ -25,5 +25,5 @@
   "name": "Vade Cloud",
   "slug": "vade-cloud",
   "uuid": "203d043e-a1ea-4b5c-81ef-6fc0f8a6a9d3",
-  "version": "1.0.1"
+  "version": "1.0.3"
```
squioc, line 28: Fix the version number.
```suggestion
  "version": "1.0.2"
```
Could you also add an entry in the CHANGELOG.md?
Thanks

squioc (changes requested):

squioc (approved):

PR #152, opened 2023-08-07
add google reports api connector

TOUFIKIzakarya:
Hi @Adamowoc, 

You will find what the google reports connector here, please take a look of it.


Adamowoc:
@TOUFIKIzakarya Could you link the issue, please ?

Google/google_module/google_drive_reports.py
```diff
@@ -0,0 +1,203 @@
...
+                maxResults=self.pagination_limit,
+                pageToken=next_key,
+                startTime=self.from_date.strftime("%Y-%m-%dT%H:%M:%SZ"),
+            )
+            .execute()
+        )
+
+        return activities
+
+    def get_repots_evts(self):
```
Adamowoc, line 155: This method is pretty big, it could be decomposed into several methods.

Google/google_module/google_drive_reports.py
```diff
@@ -0,0 +1,203 @@
...
+        now = datetime.now(timezone.utc)
+
+        self.log(message=f"Creating a google credentials objects", level="info")
+
+        activities = self.get_activities()
+
+        items = activities.get("items", [])
+        next_key = activities.get("nextPageToken")
```
Adamowoc, line 163: If the connector gets restarted, does it loose its progress and starts from the first page ?

Adamowoc (changes requested):
Could you add an entry in the changelog (like [this one](https://github.com/SEKOIA-IO/automation-library/blob/746698e4608fb627f1922f4825cfe5c3d9c9256a/Duo/CHANGELOG.md?plain=1#L23-L27) in Duo)?

Google/google_module/google_drive_reports.py
TOUFIKIzakarya: Good question, normally we stock the recent date seen in the cache. so if the connector restart in the middle of loop we will loose our progress and we will restart from the beginning. And that's because recent date is stocked just after the loop.

TOUFIKIzakarya: Yes that's what I'm thinking I will do. with sorting items before looping.

Adamowoc: @TOUFIKIzakarya Maybe they already are sorted, or they could be an option for that in the request :+1:

Google/google_module/google_drive_reports.py
```diff
@@ -0,0 +1,212 @@
...
+            grouped_data = []
+            response_next_page = self.get_next_activities(next_key)
+
+            next_page_items = response_next_page.get("items", [])
+            recent_date = next_page_items[0].get("id").get("time")
+            self.most_recent_date_seen = isoparse(recent_date)
+            grouped_data.extend(next_page_items)
+            self.log(message=f"Sending other batches of {len(grouped_data)} messages", level="info")
+            OUTCOMING_EVENTS.labels(intake_key=self.configuration.intake_key).inc(len(grouped_data))
+            self.push_events_to_intakes(events=grouped_data)
```
Adamowoc, line 171:
```suggestion
            grouped_data.extend(next_page_items)
            self.log(message=f"Sending other batches of {len(grouped_data)} messages", level="info")
            OUTCOMING_EVENTS.labels(intake_key=self.configuration.intake_key).inc(len(grouped_data))
            self.push_events_to_intakes(events=grouped_data)
            self.most_recent_date_seen = isoparse(recent_date)
```
I would set `most_recent_date_seen` after having sent the events

Google/tests/test_drive_reports.py
```diff
@@ -0,0 +1,140 @@
...
+    trigger.configuration = {
+        "admin_mail": os.environ["GOOGLE_ADMIN_MAIL"],
+        "intake_key": "intake_key",
+    }
+    trigger.log = Mock()
+    trigger.log_exception = Mock()
+    trigger.push_events_to_intakes = Mock()
+    trigger.get_reports_events()
+    resultats = [call.kwargs["events"] for call in trigger.push_events_to_intakes.call_args_list]
+    assert len(resultats[0]) != 0
```
Adamowoc, line 140:
```suggestion
    results = [call.kwargs["events"] for call in trigger.push_events_to_intakes.call_args_list]
    assert len(results[0]) != 0
```

Google/tests/test_drive_reports.py
```diff
@@ -0,0 +1,140 @@
...
+            with patch(
+                "google_module.google_drive_reports.DriveGoogleReports.get_activities", return_value=drive_response_NK
+            ):
+                with patch(
+                    "google_module.google_drive_reports.DriveGoogleReports.get_next_activities",
+                    return_value=drive_response,
+                ):
+                    trigger.get_reports_events()
+                    resultats = [call.kwargs["events"] for call in trigger.push_events_to_intakes.call_args_list]
+                    assert len(resultats[0]) != 0
```
Adamowoc, line 109:
```suggestion
                    results = [call.kwargs["events"] for call in trigger.push_events_to_intakes.call_args_list]
                    assert len(results[0]) != 0
```

Adamowoc (changes requested):

Google/google_module/google_drive_reports.py
TOUFIKIzakarya: Thanks, Good idea

Adamowoc (approved):

PR #153, opened 2023-08-10
Update logic in Crowdstrike Telemetry processing to avoid delay

vg-svitla:
1. Add prometheus metrics
2. Add tests for logs
3. Change logic of how the files in s3 get processed


Adamowoc (changes requested):
Looks good, can you update the CHANGELOG, please ?

Adamowoc (approved):
Let's try this :)

PR #214, opened 2023-08-22
Add metrics, revert some changes and so on

vg-svitla:


CrowdStrike/crowdstrike_telemetry/pull_telemetry_events.py
```diff
@@ -96,63 +94,22 @@ async def get_crowdstrike_events(self) -> list[str]:
                     validated_sqs_messages.append(CrowdStrikeNotificationSchema.parse_raw(message))
                 except Exception:
                     logger.warning("Invalid notification message")
+            if validated_sqs_messages:  # pragma: no cover
```
squioc, line 97: Maybe we can add an info log when the list of sqs messages is empty

squioc (approved):

squioc (approved):

PR #218, opened 2023-08-28
fix: Small fix in Sekoia.io feed trigger

Darkheir:
In the current implementation the trigger never sends a critical log to the api, so the trigger is never stopped.

It will start over and over again.

With the changes we made, we:
* send a `critical` error if needed
* Wait until the pod is killed

This PR also removes code that is already implemented in the SDK side:
* Method to exit
* Stop event
* Hooking on signals


squioc (approved):
LGTM

PR #226, opened 2023-08-31
Add Wait action

lvoloshyn-sekoia:


Utils/utils/__init__.py
squioc, line 1: Could you import UtilsWait in this file and register the class with the manifest?

Utils/action_utils_wait.json
```diff
@@ -0,0 +1,25 @@
+{
+    "arguments": {
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "properties": {
+            "duration": {
+                "description": "Time to wait, in seconds",
+                "type": "integer"
```
squioc, line 7: We can add [a minimum and a maximum](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-validation-keywords-for-num) to force the input of the customer
```suggestion
                "description": "Time to wait, in seconds",
                "type": "integer",
                "minimum": 0,
                "maximum": 3600
```

Utils/utils/action_utils_wait.py
```diff
@@ -0,0 +1,30 @@
...
+    Action to wait N seconds
+    """
+
+    MAX_TIME_TO_WAIT = 3600  # 1 hour
+
+    def run(self, arguments) -> None:
+        time_to_wait = arguments.get("duration")
+
+        if time_to_wait is None:
+            raise MissingActionArgumentError("duration")
```
squioc, line 19: I suggest defining a default value. However, as the json-schema required this field, it should never be none.
```suggestion
        time_to_wait = arguments.get("duration", 0)
```

Utils/utils/action_utils_wait.py
```diff
@@ -0,0 +1,30 @@
...
+        elif type(time_to_wait) != int:
+            raise ValueError("Duration should be an integer")
+
+        elif time_to_wait < 0:
+            raise ValueError("Duration can't be negative")
+
+        elif time_to_wait >= self.MAX_TIME_TO_WAIT:
+            raise ValueError("Wait duration is too long. Please don't exceed %d second(s)" % self.MAX_TIME_TO_WAIT)
+
+        sleep(time_to_wait)
```
squioc, line 30: I suggest not raising Exception but forcing `time_to_wait` to be in the range:
```suggestion

        sleep(min(self.MAX_TIME_TO_WAIT, max(time_to_wait, 0)))
```

squioc (changes requested):

squioc (approved):

PR #230, opened 2023-09-01
Crowdstrike event loop init 

vg-svitla:


CrowdStrike/crowdstrike_telemetry/pull_telemetry_events.py
```diff
@@ -212,10 +212,11 @@ async def process_s3_file(self, key: str, bucket: str | None = None) -> list[str
 
     def run(self) -> None:  # pragma: no cover
         """Runs Crowdstrike Telemetry."""
-        loop = asyncio.get_event_loop()
-
         previous_processing_end = None
+
```
squioc, line 216: I suggest having the following inside a `while self.running` loop (while keeping the other loop too).

squioc (approved):

PR #231, opened 2023-09-01
Add some logs and fix BaseException inhertance error

TOUFIKIzakarya:
- Add some logs to see if the date change or not
- Fix the `catching classes that do not inherit from BaseException is not allowed` error

### DIFF
--- a/Google/google_module/google_drive_reports.py
+++ b/Google/google_module/google_drive_reports.py
@@ -88,6 +88,8 @@ class GoogleReports(GoogleTrigger):
                 try:
                     self.get_reports_events()
 
+                except HTTPError | BaseHTTPError as ex:
+                    self.log_exception(ex, message="Failed to get next batch of events")
                 except Exception as ex:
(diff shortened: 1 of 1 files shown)


Google/google_module/google_drive_reports.py
```diff
@@ -88,10 +88,11 @@ def run(self):
                 try:
                     self.get_reports_events()
 
-                except HTTPError | BaseHTTPError as ex:
-                    self.log_exception(ex, message="Failed to get next batch of events")
```
squioc, line 92: Could you keep this block?

squioc (changes requested):

Google/google_module/google_drive_reports.py
TOUFIKIzakarya: I'm wondering if it's this line who causes the error. ok I will keep it.

squioc (approved):

PR #246, opened 2023-09-06
Darktrace: Allow unverify connections

squioc:
Add an option to not verify the server certificate when requesting over a TLS connection


TOUFIKIzakarya (approved):
LGFM

PR #262, opened 2023-09-20
Improve/feedtriggersymphony

rombernier:
- [x]  Refresh more frequently
- [x]  Use files to transfer data to actions instead of parameters
- [x]  Get revoked objects after first complete run
https://www.notion.so/Improve-Feed-trigger-in-Symphony-e69b3be40ae74c96b7134246ea4378f9?pvs=4

update sekoia.io feedtrigger and crowdstrikefalcon dissemination module

### DIFF
--- a/CrowdStrikeFalcon/crowdstrike_falcon/custom_iocs.py
+++ b/CrowdStrikeFalcon/crowdstrike_falcon/custom_iocs.py
@@ -106,7 +106,7 @@ class CrowdstrikeActionPushIOCs(CrowdstrikeActionIOC):
     def run(self, arguments):
         if arguments.get("sekoia_base_url"):
             self.sekoia_base_url = arguments.get("sekoia_base_url")
-        stix_objects = arguments.get("stix_objects", [])
+        stix_objects = self.json_argument("stix_objects", arguments)
         indicators = self.get_valid_indicators(stix_objects)
         if len(indicators["valid"]) == 0 and len(indicators["revoked"]) == 0:
             self.log("Received indicators were not valid and/or not supported")
--- a/Sekoia.io/sekoiaio/triggers/intelligence.py
+++ b/Sekoia.io/sekoiaio/triggers/intelligence.py
@@ -4,7 +4,7 @@ from threading import Event
 from urllib.parse import urljoin
 
 import requests
-from sekoia_automation.storage import PersistentJSON
+from sekoia_automation.storage import PersistentJSON, write
 from sekoia_automation.trigger import Trigger
 
 
@@ -13,9 +13,8 @@ class FeedConsumptionTrigger(Trigger):
     This trigger fetches STIX objects from Sekoia.io feed API
     """
 
-    API_URL_ADDITIONAL_PARAMETERS = ["skip_expired=true", "include_revoked=false"]
-    frequency: int = 3600  # Frequency in seconds
-    batch_size_limit: int = 200
+    API_URL_ADDITIONAL_PARAMETERS = ["skip_expired=true"]
+    frequency: int = 300  # Frequency in seconds, previous value 3600
 
     _STOP_EVENT_WAIT = 120
 
@@ -24,16 +23,29 @@ class FeedConsumptionTrigger(Trigger):
         self.context = PersistentJSON("context.json", self._data_path)
         self.next_cursor = None
         self.resume_on_errors = False
+        self.first_run = True
 
     @property
     def feed_id(self) -> str:
         return self.configuration.get("feed_id", "d6092c37-d8d7-45c3-8aff-c4dc26030608")
 
     @property
+    def to_file(self) -> bool:
+        return self.configuration.get("to_file", False)
+
+    @property
+    def batch_size_limit(self) -> int:
+        return 2000 if self.to_file else 200
+
+    @property
     def url(self):
         url = (
-            urljoin(self.module.configuration["base_url"], f"v2/inthreat/collections/{self.feed_id}/objects")
-            + f"?limit={self.__class__.batch_size_limit}"
+            urljoin(
+                self.module.configuration["base_url"],
+                f"v2/inthreat/collections/{self.feed_id}/objects",
+            )
+            + f"?limit={self.batch_size_limit}"
+            + f"&include_revoked={not self.first_run}"
         )
         if len(self.__class__.API_URL_ADDITIONAL_PARAMETERS) > 0:
             url += "&" + "&".join(self.__class__.API_URL_ADDITIONAL_PARAMETERS)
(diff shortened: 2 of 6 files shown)


Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -32,18 +35,22 @@ def feed_id(self) -> str:
...
         )
         if len(self.__class__.API_URL_ADDITIONAL_PARAMETERS) > 0:
             url += "&" + "&".join(self.__class__.API_URL_ADDITIONAL_PARAMETERS)
 
+        # depending on context, add cursor info and request revoked IOC
         with self.context as cache:
             cursor = cache.get("cursors", {}).get(self.feed_id)
             if cursor:
-                return f"{url}&cursor={cursor}"
+                return f"{url}&cursor={cursor}&include_revoked=true"
```
Darkheir, line 51: This will include the revoked indicator not after the first full sync but after we fetch the first page.

Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -100,9 +107,24 @@ def next_batch(self):
...
+            file_path = str(stix_objects_path.relative_to(work_dir))
+            # Attach the file to the event and send objects as result of the trigger
+            event_name = f"Sekoia.io feed - batch of {len(objects)} objects"
             self.send_event(
-                event_name=f"Sekoia.io feed - batch of {len(objects)} objects", event={"stix_objects": objects}
+                event_name=event_name,
+                event=dict(event_path=file_path),
+                directory=directory,
+                remove_directory=True,
             )
```
Darkheir, line 128: This is a breaking change that will break all the existing playbooks expecting to find a `stix_objects` attribute and not a path

Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -100,9 +107,24 @@ def next_batch(self):
...
+            work_dir = self._data_path.joinpath("stix_objects").joinpath(str(uuid.uuid4()))
+            work_dir.mkdir(parents=True, exist_ok=True)
+
+            stix_objects_path = work_dir.joinpath("stix_objects.json")
+            with stix_objects_path.open("w") as fp:
+                fp.write(orjson.dumps(objects).decode("utf-8"))
+
+            # Send event
+            directory = str(work_dir.relative_to(self._data_path))
+            file_path = str(stix_objects_path.relative_to(work_dir))
```
Darkheir, line 120: In the SDK there is an helper `storage.write` doing that.

Darkheir (changes requested):
Those changes in the trigger will break existing playbooks.

To make the change with breaking we must introduce a new attribute `to_file` that should be false by default allowing the user to choose how to sync the data.

Also your limit to 2000 will break existing playbooks because the size of the event will be too big to be processed by actions.

Finally the version in the manifest should be bumped.

Darkheir:
> Thanks for your review, I will make the changes, what is the size limit of event to be processed by actions ?

I think that  if the client wants to write it in a file you can set the limit to 2000 else keep it at 200

CrowdStrikeFalcon/crowdstrike_falcon/custom_iocs.py
```diff
@@ -106,7 +106,10 @@ def get_valid_indicators(self, stix_objects):
     def run(self, arguments):
         if arguments.get("sekoia_base_url"):
             self.sekoia_base_url = arguments.get("sekoia_base_url")
-        stix_objects = arguments.get("stix_objects", [])
+        if "file_path" in arguments:
```
Darkheir, line 109: The action never expects a `file_path` arguments. The action manifest should be updated to have:

* a `stix_objects_path` arguments
* Either `stix_objects` or `stix_objects_path` required

Then replacing `stix_objects = arguments.get("stix_objects", [])` with `stix_objects = self.json_argument("stix_objects", arguments)` will be enough to look inside the file or the argument.

The version in the manifest must also be bumped

Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -24,16 +26,34 @@ def __init__(self, *args, **kwargs):
         self.context = PersistentJSON("context.json", self._data_path)
         self.next_cursor = None
         self.resume_on_errors = False
+        self.first_run = True
 
     @property
     def feed_id(self) -> str:
         return self.configuration.get("feed_id", "d6092c37-d8d7-45c3-8aff-c4dc26030608")
 
+    @property
```
Darkheir, line 35: You must update the trigger manifest to include the `to_file` argument

Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -24,16 +26,34 @@ def __init__(self, *args, **kwargs):
...
+        return self.configuration.get("to_file", False)
+
+    @property
+    def batch_size_limit(self) -> int:
+        file_batch_size_limit: int = 2000  # limit to 2000 (the api limit) for event send by file
+        url_batch_size_limit: int = 200  # limit to 200 for event send by url parameters
+        if self.configuration.get("to_file", False):
+            return file_batch_size_limit
+        else:
+            return url_batch_size_limit
```
Darkheir, line 46:
```suggestion
        return 2000 if self.to_file else 200
```

Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -100,10 +120,17 @@ def next_batch(self):
...
-            self.send_event(
-                event_name=f"Sekoia.io feed - batch of {len(objects)} objects", event={"stix_objects": objects}
-            )
+            event_name = f"Sekoia.io feed - batch of {len(objects)} objects"
+            if self.to_file:
+                filepath = write("stix_objects.json", {"stix_objects": objects})
+                directory = filepath.parent.as_posix()
+
+                self.send_event(
+                    event_name=event_name, event={"file_path": filepath}, directory=directory, remove_directory=True
```
Darkheir, line 129: The convention is to use the same name as for when it is not in a file with the `_path` suffix.

```suggestion
                    event_name=event_name, event={"stix_objects_path": filepath}, directory=directory, remove_directory=True
```

You must also update the trigger manifest to add this attribute as a result

Darkheir (changes requested):
The version in the module manifest must also be bumped

rombernier:
Hello @Darkheir, I didn't forget to bump the manifest version this time. I bump the middle number as it is a new functionality and not a bugfix (#sumver) , can you check the PR once more ? Thanks

Darkheir (approved):
Code looks good to me.

@squioc Do you want someone from your team to also validate the PR ?

rombernier:
Hello @squioc, can you look and validate this PR, to be able to check on test app ? Thanks

squioc (approved):
LGTM

rombernier:
Thanks for the review @squioc, do you have a trick to merge PR with check stuck in status in progress or codecov failed ?

rombernier:
Hello @Darkheir the build of this PR merge failed on Test module CrowdStrikeFalcon on mypy step : https://github.com/SEKOIA-IO/automation-library/actions/runs/6379736513/job/17386346337
but the test on the branch where fine, do you know if there is an issue that can create this problem ? I need to create an other PR to fix this ? Thanks

Darkheir:
@rombernier This is not a big issue, just a typing one. 

Instead of `def get_retry_after(self, response: HTTPResponse) -> float | None:` the method should have the following signature `def get_retry_after(self, response: BaseHTTPResponse) -> float | None:`.

IMO You can leave it as it is

PR #274, opened 2023-09-25
Fix incorrect logging

lvoloshyn-sekoia:


squioc (changes requested):
Can you update the version number of the module and add an entry in the CHANGELOG.md?

squioc (approved):

PR #278, opened 2023-09-26
Connector to work with Azure Blob

vg-svitla:


AzureBlob/trigger_azure_blob_storage_events.json
```diff
@@ -0,0 +1,30 @@
...
+    "arguments": {
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "title": "Trigger Azure Blob Storage configuration",
+        "type": "object",
+        "properties": {
+            "intake_server": {
+                "description": "Server of the intake server (e.g. 'https://intake.sekoia.io')",
+                "default": "https://intake.sekoia.io",
+                "type": "string"
+            },
```
squioc, line 11: Still needed :(
```suggestion
            },
            "chunk_size": {
              "type": "integer",
              "description": "The max size of chunks for the batch processing",
              "default": 1000
            },
```

squioc (changes requested):
Good job.
Just a parameter missing from the configuration schema.

squioc (changes requested):
Could you move the trigger in the existing azure module?
Thanks.

Azure/azure_helpers/storage.py
```diff
@@ -0,0 +1,97 @@
...
+                file_name = str(file.name)
+                async for chunk in stream.chunks():
+                    await file.write(chunk)
+
+                return file_name, None
+
+        # This operation is blocking
+        result = await stream.readall()
+
+        return None, result
```
squioc, line 97: Why not create an abstract class for blob and having two implementations (one for files and another one for raw content), return one of this implementation from the method and handle differences in these classes?

squioc (approved):

Azure/azure_helpers/storage.py
vg-svitla: To be honest, it is a bit difficult for further support and lets say someone will try to refactor/update logic - two different class implementation will confuse. Also there might be situation when we would like to use different method based on some flag or config - does it mean that we want to initialise both of them? 
I'm fine that we return always tuple with optional filename and optional bytes...
If you would like to split it - I'm also fine with that

vg-svitla:
> **codecov/project ** — 90.21% (-0.01%) compared to db18df7

Seems like problem relates to codecov and not black..

PR #280, opened 2023-09-27
Update some connectors to use AsyncConnector

vg-svitla:


squioc (changes requested):
Thanks for the PR.

For each module, could you bump the version number in the manifest.json and add an entry in the CHANGELOG.md?

squioc (approved):

PR #345, opened 2023-10-05
Refactoring of AWS module

vg-svitla:


AWS/aws_helpers/sqs_wrapper.py
```diff
@@ -0,0 +1,131 @@
...
+
+                logger.info(f"Received {len(result)} messages from sqs queue {self._configuration.queue_name}")
+
+                yield result
+            finally:
+                # We should delete messages from queue after releasing context manager if it is configured
+                if delete_consumed_messages and response.get("Messages", []):
+                    logger.info("Deleting consumed messages from sqs")
+                    for message in response.get("Messages", []):
+                        await sqs.delete_message(QueueUrl=queue_url, ReceiptHandle=message["ReceiptHandle"])
```
squioc, line 131: Does the async sqs client not allow messages to be deleted in batch? Like [this code](https://github.com/SEKOIA-IO/automation-library/blob/main/AWS/aws/sqs/trigger_sqs_messages.py#L64-L67)

AWS/connectors/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+        loop = asyncio.get_event_loop()
+
+        previous_processing_end = None
+        try:
+            while True:
+                processing_start = time.time()
+                if previous_processing_end is not None:
+                    EVENTS_LAG.labels(intake_key=self.configuration.intake_key).observe(
+                        processing_start - previous_processing_end
+                    )
```
squioc, line 76: The SQS [ReceiveMessage](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html) action allow you to get the `SentTimestamp` attribute name. I wonder if using this attribute to calculate the delay is more appropriate. Could you take a look?
Thanks

AWS/connectors/trigger_sqs_messages.py
```diff
@@ -0,0 +1,68 @@
...
+from sekoia_automation.connector import DefaultConnectorConfiguration
+
+from aws_helpers.sqs_wrapper import SqsConfiguration, SqsWrapper
+from connectors import AbstractAwsConnector
+
+
+class AwsSqsMessagesTriggerConfiguration(DefaultConnectorConfiguration):
+    """Contains configuration for AwsSqsMessagesTrigger."""
+
+    frequency: int = 60
```
squioc, line 14: This parameter seems to not be used in the connector? Is useless or a `asyncio.sleep` is missing?

squioc (commented):
Thanks for the job.

I got some questions.

AWS/connectors/trigger_sqs_messages.py
vg-svitla: Yep! Thank you! Fixed

AWS/aws_helpers/sqs_wrapper.py
vg-svitla: As per docs from here https://github.com/aio-libs/aiobotocore#supported-aws-services `sqs` wrapper is only partially supported.
I was not able to find any tests that  proves that delete batch operation works as expected so I tried myself.
I was not able to use it, as I faced generic `Exception` with strange message.
I can investigate it more, or we can try to use custom implementation

AWS/connectors/__init__.py
vg-svitla: If I'm correct the main goal of `EVENTS_LAG` is to check delay between runs in connector.
As per official docs `SentTimestamp` is the time when the message was sent to the queue.
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html
Lets discuss in DM

squioc (approved):

Issue #346, opened 2023-10-09
Implement an authorize and a drop list on AWS Cloudtrail


gbossert:
Create a filter for all the Cloudtrail events

- [ ] When the `eventSource` is `s3.amazonaws.com`, only events with the following eventName must be collected:

- Delete*
- Create*
- Copy*
- Put*
- Restore*
- GetObject
- GetObjectTorrent
- ListBuckets

- [ ] When the `eventSource` is **not** `s3.amazonaws.com`, do not collect events with the following eventName:
- List*
- Describe*

PR #347, opened 2023-10-09
Add logic based to CloudtrailLogs parser on #346

vg-svitla:
https://github.com/SEKOIA-IO/automation-library/issues/346
Here is a context

### DIFF
--- a/AWS/aws/s3/base.py
+++ b/AWS/aws/s3/base.py
@@ -58,7 +58,7 @@ class AWSS3Worker(Thread, metaclass=ABCMeta):
     def log(self, *args, **kwargs):
         self.trigger.log(*args, **kwargs)
 
-    def log_exception(self, *args, **kwargs):
+    def log_exception(self, *args, **kwargs):  # pragma: no cover
         self.trigger.log_exception(*args, **kwargs)
 
     def send_records(self, records: list, event_name: str):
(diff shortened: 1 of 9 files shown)


gbossert:
Thank you Vladyslav, it looks almost good to me. The only remarks I have:

- you use the in operator : `x in payload.get("eventName")` but very most of the filters are "startswith"
- can you add some small tests
- is it possible to move this code to the dedicated cloudtrail trigger so it doesn't apply to all the S3 trigger ?

gbossert:
@vg-svitla thanks for the tests.

I think the filtering for other services than S3 is inverted.
It is a drop-list, meaning that events that match should NOT be collected.

For example the following two method calls should return False

```
trigger.is_valid_payload({"eventSource": "random.amazonaws.com", "eventName": "List"})
trigger.is_valid_payload({"eventSource": "random.amazonaws.com", "eventName": "Describe"})
```

While the following should return True

```
trigger.is_valid_payload({"eventSource": "random.amazonaws.com", "eventName": "Delete"})
trigger.is_valid_payload({"eventSource": "random.amazonaws.com", "eventName": "Create"})
trigger.is_valid_payload({"eventSource": "random.amazonaws.com", "eventName": "Random"})
```

gbossert (approved):

PR #350, opened 2023-10-11
Add action search reports

TOUFIKIzakarya:
Add get reports action :
- Modify main.py
- Modify CHANGELOG.md
- Modify __init__.py
- Add action json


TOUFIKIzakarya:
As we have done another PR for that, and we already have search action in our module, i will close this PR.

PR #355, opened 2023-10-17
Add TrendMicro connector

lvoloshyn-sekoia:


TrendMicroEmailSec/trendmicro_modules/client/auth.py
squioc, line 1: Why not use [HTTPBasicAuth](https://docs.python-requests.org/en/latest/user/authentication/#basic-authentication) from python-requests instead?

TrendMicroEmailSec/trendmicro_modules/trigger_email_sec.py
```diff
@@ -0,0 +1,256 @@
...
+    @property
+    def frequency(self):
+        return self.connector.configuration.frequency
+
+    @cached_property
+    def client(self) -> ApiClient:
+        auth = TrendMicroAuthentication(
+            username=self.connector.module.configuration.username,
+            api_key=self.connector.module.configuration.api_key,
+        )
```
squioc, line 53: I think that [HTTPBasicAuth](https://docs.python-requests.org/en/latest/user/authentication/#basic-authentication) is more suitable.

TrendMicroEmailSec/trendmicro_modules/trigger_email_sec.py
```diff
@@ -0,0 +1,270 @@
...
+        # if undefined, retrieve events from the last 5 minutes
+        if most_recent_ts_seen is None:
+            return now - 5 * 60
+
+        # We can't retrieve messages older than 72 hours
+        one_week_ago = now - 72 * 60 * 60
+        if most_recent_ts_seen < one_week_ago:
+            most_recent_ts_seen = one_week_ago
+
+        return most_recent_ts_seen
```
squioc, line 75: Why do you use timestamp? `datetime` wasn't suitable?

squioc (approved):
LGTM.

TrendMicroEmailSec/trendmicro_modules/trigger_email_sec.py
lvoloshyn: We can use `datetime` and `timedelta` here and anywhere in the connector as well. But IMO it's just more concise and aligned with other connectors this way

PR #356, opened 2023-10-17
feat(sekoia.io): Better handling of live url

Darkheir:
Improve live URL computation to support several regions.

### DIFF
--- a/Sekoia.io/sekoiaio/triggers/base.py
+++ b/Sekoia.io/sekoiaio/triggers/base.py
@@ -7,6 +7,7 @@ patch_all()
 import json
 import random
 import time
+from urllib.parse import urlparse
 
 import requests
 from sekoia_automation.trigger import Trigger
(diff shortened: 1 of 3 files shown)


Sekoia.io/sekoiaio/triggers/base.py
```diff
@@ -55,10 +56,13 @@ def liveapi_url(self):
...
             base_url = self.module.configuration["base_url"]
-            if "app.test.sekoia.io" in base_url:
+            domain = urlparse(base_url).netloc
+            if domain in ["app.test.sekoia.io", "api.test.sekoia.io"]:
                 liveapi_url = "wss://live.test.sekoia.io/"
-            else:
+            elif domain in ["app.sekoia.io", "api.sekoia.io"]:
                 liveapi_url = "wss://live.sekoia.io/"
+            else:
+                liveapi_url = f"wss://{domain}/live"
```
otetard, line 65: `wss://app.sekoia.io/live` and `wss://app.test.sekoia.io/live` should work fine. Maybe we could fallback on `wss://{domain}/live` in all situations.

PR #370, opened 2023-10-18
move conf from account to trigger for crowdstrike

rombernier:


rombernier:
Hello @squioc, I forgot to move conf from account to trigger for crowdstrike falcon module in the previous PR, can you check and approved ? thanks

CrowdStrikeFalcon/manifest.json
squioc, line 1: Could you update the number version of the manifest?

squioc (changes requested):

CrowdStrikeFalcon/manifest.json
rombernier: Right, I forgot that, I don't have the reflex yet, thanks

squioc (approved):

PR #371, opened 2023-10-18
fix for looping authentification

Adamowoc:


squioc (changes requested):
Could you update the version number in the manifest?

Be careful!! There is already another [PR about CrowdStrike](https://github.com/SEKOIA-IO/automation-library/pull/370).

a-leonardi (approved):

squioc (approved):

PR #373, opened 2023-10-19
fix module sekoiaio feedioc

rombernier:


Darkheir (commented):
Could you also bump the module version in the manifest ?

rombernier:
> Could you also bump the module version in the manifest ?

Right, I forgot that, doing it now

Darkheir (approved):

PR #377, opened 2023-10-23
update current log

rombernier:


CrowdStrikeFalcon/crowdstrike_falcon/custom_iocs.py
```diff
@@ -110,7 +110,7 @@ def run(self, arguments):
         stix_objects = self.json_argument("stix_objects", arguments)
         indicators = self.get_valid_indicators(stix_objects)
         if len(indicators["valid"]) == 0 and len(indicators["revoked"]) == 0:
-            self.log("Received indicators were not valid and/or not supported")
+            self.log(f"Received indicators were not valid and/or not supported in stix_objects {str(stix_objects)}")
```
squioc, line 113: I'm afraid that `stix_object` will be huge for the logger.
Maybe the following code will be more suitable:
```suggestion
            self.log("Received indicators were not valid and/or not supported in stix_objects", level="error", stix_object=stix_objects)
```

squioc: @Darkheir What do you think about the original proposal and the suggestion? Thanks

rombernier: Ok, I commit your suggestion

Darkheir:
Is it really needed to log all the objects ? I don't know how often this log will be produced but It may generate a lot of content

rombernier:
> Is it really needed to log all the objects ? I don't know how often this log will be produced but It may generate a lot of content

It is useful to see the structure of the object and understand why we have this issue but if you know an other way to get the object structure we can used it ?

Darkheir:
> It is useful to see the structure of the object and understand why we have this issue but if you know an other way to get the object structure we can used it ?

The thing is the objects are going to be sent to sentry, but you will have no way of seeing them in the UI.

rombernier:
> The thing is the objects are going to be sent to sentry, but you will have no way of seeing them in the UI.
How do you debug in this case ?

Darkheir:
You should run `poetry lock` inside the directory and add the `poetry.lock` file to the PR :)

squioc (approved):
LGTM

rombernier:
> LGTM

Thanks for the approval, can you merge, I can't because of codecov/patch

PR #387, opened 2023-10-25
Exclude GetRecords and GetObjects

gbossert:


squioc (changes requested):
Could you update the version number in the manifest? Thanks.

gbossert:
I've bumped the version in the manifest and created an entry in the changelog

squioc (approved):

PR #441, opened 2023-10-29
Introduce event aggregation on AWS

gbossert:
This PR introduces the notion of event aggregation.

Events that share a comment fingerprint will be aggregated under a TTL strategy.

### DIFF
--- a/AWS/aws/event_aggregator.py
+++ b/AWS/aws/event_aggregator.py
@@ -1,7 +1,9 @@
+import copy
 import datetime
 import threading
 import time
-from typing import Callable
+
+from typing import Callable, Tuple
 from ciso8601 import parse_datetime_as_naive
 
 from pydantic import BaseModel
@@ -10,19 +12,23 @@ from pydantic import BaseModel
 class Fingerprint(BaseModel):
     condition_func: Callable[[dict], bool]
     build_hash_str_func: Callable[[dict], str]
+    ttl: int
 
 
 class Aggregation(BaseModel):
     start: datetime.datetime
     end: datetime.datetime
+    ttl: int
     count: int
     event: dict
 
     def get_aggregated_event(self):
-        aggregated_event = self.event
-        aggregated_event["event"]["start"] = self.start.isoformat()
-        aggregated_event["event"]["end"] = self.start.isoformat()
-        aggregated_event["event"]["count"] = self.count
+        aggregated_event = copy.deepcopy(self.event)
+        aggregated_event["sekoiaio"]["repeat"] = {
+            "start": self.start.isoformat(),
+            "end": self.end.isoformat(),
+            "count": self.count,
+        }
         return aggregated_event
 
 
@@ -34,14 +40,12 @@ class EventAggregatorTTLThread(threading.Thread):
     def __init__(
         self,
         event_aggregator: "EventAggregator",
-        ttl: int,
         on_flush_func: Callable[[dict], None],
         delay: float = 10,
     ):
         super().__init__()
         self.event_aggregator = event_aggregator
         self.f_must_stop = False
-        self.ttl = ttl
         self.on_flush_func = on_flush_func
         self.delay = delay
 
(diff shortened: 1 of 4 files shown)


AWS/aws/event_aggregator.py
```diff
@@ -0,0 +1,169 @@
...
+class Aggregation(BaseModel):
+    start: datetime.datetime
+    end: datetime.datetime
+    count: int
+    event: dict
+
+    def get_aggregated_event(self):
+        aggregated_event = self.event
+        aggregated_event["event"]["start"] = self.start.isoformat()
+        aggregated_event["event"]["end"] = self.start.isoformat()
```
Darkheir, line 24:
```suggestion
        aggregated_event["event"]["end"] = self.end.isoformat()
```

AWS/aws/event_aggregator.py
```diff
@@ -0,0 +1,169 @@
...
+            event_hash = self.get_hash(event)
+            # if no hash can be computed, we don't aggregate and forward the event
+            if not event_hash:
+                return event
+
+            # the aggregation stores the parsed timestamp
+            event_dt = parse_datetime_as_naive(event["@timestamp"])
+
+            # if hash is already known
+            with self.lock:
```
Darkheir, line 145: So if I understand correctly, first occurrence is sent immediately and the aggregated one will follow once TTL is reached. Is that right ?

AWS/aws/event_aggregator.py
```diff
@@ -0,0 +1,169 @@
...
+class EventAggregator:
+    ttl_thread: EventAggregatorTTLThread | None
+    lock: threading.Lock
+
+    def __init__(self, aggregation_definitions: dict[str, list[Fingerprint]]):
+        self.aggregation_definitions = aggregation_definitions
+        self.aggregations: dict[str, Aggregation] = dict()
+        self.lock = threading.Lock()
+
+    def start_flush_on_ttl(self, ttl: int, on_flush_func: Callable[[dict], None], delay: float = 10):
```
Darkheir, line 71: TTL is global and can't be changed depending on the kind of event. 

I'm totally fine with it, but I'm wondering if we can have cases where some aggregations should be forwarded faster than other ?

AWS/aws/event_aggregator.py
```diff
@@ -0,0 +1,169 @@
...
+        with self.lock:
+            for event_hash, aggregation in self.aggregations.items():
+                if aggregation.start + datetime.timedelta(seconds=ttl) < datetime.datetime.utcnow():
+                    event_hashes_to_flush.append(event_hash)
+
+                    if aggregation.count > 0:
+                        aggregated_events.append(aggregation.get_aggregated_event())
+
+            for event_hash in event_hashes_to_flush:
+                del self.aggregations[event_hash]
```
Darkheir, line 119: Could be done in the first iteration since we iterate on `.items()`

Darkheir (commented):
Code logic looks good.

It would be nice to have it in the SDK so all the connectors may benefit of this capability

AWS/aws/event_aggregator.py
gbossert: yes exactly

AWS/aws/event_aggregator.py
gbossert: well spotted, I will upgrade the aggregation definition to support custom ttl

AWS/aws/event_aggregator.py
gbossert: ok I didn't know this method was compatible with dict edition. I will simplify accordingly.

gbossert:
This PR is moved in the SDK : https://github.com/SEKOIA-IO/sekoia-automation-sdk/pull/87

AWS/aws/event_aggregator.py
gbossert: .items() doesnt seems compatible with inplace edition of dict. I used list(.keys()) instead

PR #444, opened 2023-10-30
Fix get_context bug

TOUFIKIzakarya:
. Fix the condition in the run function


squioc (approved):
Can you add an entry in the CHANGELOG.md? Thanks

squioc (approved):

PR #463, opened 2023-11-03
PandaSecurity: Additional custom header to platform requests

vg-svitla:


PandaSecurity/aether_endpoint_security_api/trigger_security_events.py
```diff
@@ -2,6 +2,7 @@
 import time
 from collections.abc import Generator, Sequence
 from datetime import datetime, timedelta
+from functools import cached_property
```
squioc, line 5: This import seems useless

squioc (changes requested):

PR #466, opened 2023-11-03
Some fixes after testing

TOUFIKIzakarya:


AzureActiveDirectory/azure_ad/base.py
```diff
@@ -98,7 +102,7 @@ def validate_id_or_userPrincipalName(cls, values):
 
 class RequiredTwoUserArguments(SingleUserArguments):
     userNewPassword: str | None = Field(
-        None,
+        secret=True,
```
squioc, line 105: Are you sure that secrets work with arguments?

TOUFIKIzakarya: Hi @squioc, It works for `client_secret` and `password` fields for `AzureADConfiguration`. I think it's the some thing

squioc: `secret` only work for configuration, not arguments. Can you fix this code? Thanks?

squioc (changes requested):

squioc (approved):
LGTM

PR #473, opened 2023-11-06
Httpresponse

Adamowoc:
MyPy is annoying


squioc (changes requested):
Could you increase the version number in the manifest?

Darktrace/darktrace_modules/client/retry.py
```diff
@@ -1,6 +1,5 @@
 import time
```
otetard, line 1: You can also move to `urllib3` 2.x that includes `BaseHTTPResponse`. You can add an explicit dependency on that version via `poetry add urllib3@latest`

Adamowoc: Thank you for the advice @otetard, I will try that.

squioc (approved):

PR #475, opened 2023-11-08
fix last_ts for unordered aianalyst events

Adamowoc:


squioc (changes requested):
Could you increase the version number in the manifest? Thanks

squioc (approved):

PR #483, opened 2023-11-13
fix file path in trigger

rombernier:


Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -114,10 +114,10 @@ def next_batch(self):
 
             event_name = f"Sekoia.io feed - batch of {len(objects)} objects"
             if self.to_file:
-                filepath = write("stix_objects.json", {"stix_objects": objects})
+                filepath = write(self.__class__.FILE_NAME, {"stix_objects": objects})
```
Darkheir, line 117: I think you should directly save the list of objects or the next actions will have a dict instead of an expected list.
Also `__class__` is not needed

```suggestion
                filepath = write(self.FILE_NAME, objects)
```

Sekoia.io/sekoiaio/triggers/intelligence.py
```diff
@@ -114,10 +114,10 @@ def next_batch(self):
...
                 self.send_event(
                     event_name=event_name,
-                    event={"stix_objects_path": str(filepath)},
+                    event={"stix_objects_path": self.__class__.FILE_NAME},
```
Darkheir, line 120:
```suggestion
                    event={"stix_objects_path": self.FILE_NAME},
```

Darkheir (changes requested):
Module version should be bumped

Darkheir (approved):

PR #511, opened 2023-11-20
Fix/thehive

rombernier:
Fix error with some alert with a long type like: intrusion/application-compromise
Alert type limit to 32 char in TheHive API: https://docs.strangebee.com/thehive/api-docs/#operation/Create%20Alert

Also add link to sekoia alert in description, because external link not visible on TheHive GUI

Fix and updates requested by client

### DIFF
--- a/TheHive/thehive/create_alert.py
+++ b/TheHive/thehive/create_alert.py
@@ -96,7 +96,7 @@ class TheHiveCreateAlert(Action):
 
         alert_type = f"{arg_alert['alert_type']['category']}/{arg_alert['alert_type']['value']}"
         if len(alert_type) > 32:
-            alert_type = arg_alert['alert_type']['category'][:32]  # limit to 32 char, max of thehive api
+            alert_type = arg_alert["alert_type"]["category"][:32]  # limit to 32 char, max of thehive api
         link = f"https://app.sekoia.io/operations/alerts/{arg_alert['short_id']}"
         alert: Alert = Alert(
             id=arg_alert["uuid"],
@@ -104,7 +104,7 @@ class TheHiveCreateAlert(Action):
             date=arg_alert["created_at"],
             tags=[],
             title=arg_alert["title"],
-            type=alert_type
+            type=alert_type,
             source="Sekoia.io",
             sourceRef=arg_alert["short_id"],
             externalLink=link,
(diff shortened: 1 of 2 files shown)


TheHive/thehive/create_alert.py
```diff
@@ -94,17 +94,19 @@ def run(self, arguments: dict[str, Any]):
                             )
                         )
 
+        alert_type = f"{arg_alert['alert_type']['category']}/{arg_alert['alert_type']['value']}"
+        link = f"https://app.sekoia.io/operations/alerts/{arg_alert['short_id']}"
```
Darkheir, line 98: Will not work with other regions than FRA1

TheHive/thehive/create_alert.py
```diff
@@ -94,17 +94,19 @@ def run(self, arguments: dict[str, Any]):
                             )
                         )
 
+        alert_type = f"{arg_alert['alert_type']['category']}/{arg_alert['alert_type']['value']}"
```
Darkheir, line 97: Maybe try to cut it in a way that will be less strange in the UI ?

```suggestion
        alert_type = f"{arg_alert['alert_type']['category']}/{arg_alert['alert_type']['value']}"
        if len(alert_type) > 32:
            alert_type = arg_alert['alert_type']['category'][:32]  # limit to 32 char, max of thehive api
```

TheHive/thehive/create_alert.py
```diff
@@ -94,17 +94,19 @@ def run(self, arguments: dict[str, Any]):
...
+        link = f"https://app.sekoia.io/operations/alerts/{arg_alert['short_id']}"
         alert: Alert = Alert(
             id=arg_alert["uuid"],
             severity=arg_alert["urgency"]["severity"] // 25 + 1,  # from 0-100 to 1-4
             date=arg_alert["created_at"],
             tags=[],
             title=arg_alert["title"],
-            type=f"{arg_alert['alert_type']['category']}/{arg_alert['alert_type']['value']}",
+            type=alert_type[:32],  # limit to 32 char, max of thehive api
```
Darkheir, line 105:
```suggestion
            type=alert_type
```

TheHive/thehive/create_alert.py
rombernier: Indeed, can we keep it like that for now,
How this type of issue is managed in other module ?

Darkheir: Yes we can keep it this way for now. 

I think we will need to address the issue with changes in our API so the action can know in which region they are running.

Darkheir (approved):

PR #512, opened 2023-11-20
fix issue with alert date sec instead of ms

rombernier:
date in ms for the hive: https://docs.strangebee.com/thehive/api-docs/#operation/Create%20Alert
date in sec for module sekoia get alert


Darkheir (approved):
Nice catch

PR #514, opened 2023-11-22
add historicmodelonly=false

Adamowoc:
Adding historicmodelonly=false to unify the clients logs


squioc (changes requested):
Could you update the version number in manifest.yml and add an entry in the changelog? Thanks.

squioc (approved):

PR #530, opened 2023-11-24
Bugfix: Checkpoint fix datetime parsing problem

vg-svitla:


Checkpoint/client/http_client.py
```diff
@@ -81,20 +81,60 @@ async def get_harmony_mobile_alerts(
...
+        Also, sometimes it can be `No data` value.
+
+        Args:
+            value: str
+
+        Returns:
+            datetime | None:
+        """
+        if value is None:
+            return None
```
squioc, line 128: Should we not add a test about `No data` and return `None` in this case?

squioc (approved):

PR #542, opened 2023-11-28
Checkpoint: fix naive datetimes

squioc:
Convert naive datetimes from events into aware ones


TOUFIKIzakarya (commented):
There's a test file `test_http_client.py` who fails

TOUFIKIzakarya (approved):

PR #614, opened 2023-12-01
update sekoia and crowdstrike module to keep mode file only

rombernier:
Keep file mode only to simplify, will this is not used by clients we can make the update


squioc (commented):
Do the argument `stix_object` used?

rombernier:
> Do the argument `stix_object` used?

Not in file mode

CharlesLR-sekoia:
Hello @squioc any modification to be performed to complete this review?

rombernier:
@squioc, test are ok, can you check/approved and merged :) Thanks

squioc (approved):

Darkheir:
**This change is going to break multiple playbooks that expect `stix_objects` as an output and not `stix_objects_path`.**

Are you OK with breaking clients playbooks using those triggers ?

rombernier:
> **This change is going to break multiple playbooks that expect `stix_objects` as an output and not `stix_objects_path`.**
> 
> Are you OK with breaking clients playbooks using those triggers ?

Who do you check which clients are using the action ? For me clients are using the trigger only: 
https://grafana.dashboard.sekoia.io/d/c4e8b67d-16d8-423b-91cd-641acb820df2/modules?orgId=1&var-module_name=CrowdStrike%20Falcon&var-module_uuid=5948a752-5c16-4bbd-9b12-3bc907ee05cf
But I'm not sure about this grafana

rombernier:
> **This change is going to break multiple playbooks that expect `stix_objects` as an output and not `stix_objects_path`.**
> 
> Are you OK with breaking clients playbooks using those triggers ?

How do you know which clients are using this triggers ? With this dashboard https://grafana.dashboard.sekoia.io/d/c4e8b67d-16d8-423b-91cd-641acb820df2/modules?orgId=1&var-module_name=Sekoia.io&var-module_uuid=ac6100ed-3fb7-4355-83ac-049c14aa44fd I can't find clients, but I'm not sur about this dashboard

Darkheir:
I listed the triggers configurations using those triggers and checked the associated playbooks from the DB

Darkheir:
Regarding active playbooks I think that only Mirakl is concerned

Darkheir:
If you want we can deploy the new version of the module and immediately update the playbooks so they won't break. Is it ok for you this way ?

rombernier:
> If you want we can deploy the new version of the module and immediately update the playbooks so they won't break. Is it ok for you this way ?

Yes, it's ok for me

PR #616, opened 2023-12-05
Fix an exception in CrowdStrike Falcon

lvoloshyn-sekoia:


CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
```diff
@@ -220,7 +220,8 @@ def run(self) -> None:
                 while not self.f_stop.is_set():
                     if datetime.utcnow() >= next_refresh_at:
                         self.refresh_stream(refresh_url=self.stream_info["refreshActiveSessionURL"])
-                        next_refresh_at = datetime.utcnow() + timedelta(minutes=25)
+                        time_to_refresh = int(self.stream_info["refreshActiveSessionInterval"])
+                        next_refresh_at = datetime.utcnow() + timedelta(seconds=time_to_refresh)
```
squioc, line 224: :+1:

I think you should also change the line 197.

CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
```diff
@@ -311,12 +313,15 @@ def __init__(self, *args, **kwargs) -> None:
 
         self.auth_token = None
 
-        self.app_id = f"sio-{time.time()}"
+        self.app_id = self.generate_app_id()
```
squioc, line 316: Can you move the generation of the app_id into `EventStreamReader` instead? Thanks

squioc (changes requested):
Thanks for the fixes.
Some changes to apply.

CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
lvoloshyn-sekoia: But then we won't be able to access it directly from the connector

squioc (approved):

PR #618, opened 2023-12-06
Use a single asyncio loop

lvoloshyn-sekoia:


Azure/connectors/trigger_azure_eventhub.py
```diff
@@ -139,12 +139,12 @@ async def handle_exception(self, partition_context: PartitionContext, exception:
     def run(self) -> None:  # pragma: no cover
         self.log(message="Azure EventHub Trigger has started", level="info")
 
+        loop = asyncio.new_event_loop()
+        loop.add_signal_handler(signal.SIGINT, lambda: asyncio.create_task(self.stop()))
+        loop.add_signal_handler(signal.SIGTERM, lambda: asyncio.create_task(self.stop()))
+
```
squioc, line 145: For async connector, we use [this double while](https://github.com/SEKOIA-IO/automation-library/blob/38c33968d79da4fff4a95c5ff6a1bc08764ccc23/Salesforce/salesforce/connector.py#L148-L154) to always have an asyncio loop operational.
So, I fear that putting the loop outside the `while` will lead to another issue.

vg-svitla:
Looks good to me. 
What do you think @squioc ?

vg-svitla (approved):

Azure/connectors/trigger_azure_eventhub.py
```diff
@@ -70,14 +70,6 @@ def __init__(self, *args: Any, **kwargs: Optional[Any]) -> None:
         self._stop_event = Event()
         self._consumption_max_wait_time = int(os.environ.get("CONSUMER_MAX_WAIT_TIME", "600"), 10)
 
-    async def stop(self, *args: Any, **kwargs: Optional[Any]) -> None:
-        self.log(message="Stopping Azure EventHub connector", level="info")
-        # Exit signal received, asking the processor to stop
-        self._stop_event.set()
-
-        # Close the azure EventHub client
-        await self.client.close()
```
squioc, line 79: Arf. We will not properly close the client. It may cause some issues in the consumption of the event hub.

squioc (changes requested):
How can we close the client in an async way when the `run` method is not async?
@vg-svitla @lvoloshyn-sekoia

Azure/connectors/trigger_azure_eventhub.py
lvoloshyn-sekoia: I don't believe we're actually calling `stop()` anywhere. I overrode the method in order to use callbacks for SIGINT/SIGTERM. But since we use an existing loop from AsyncConnector, we don't have to set those up.

squioc (approved):

PR #620, opened 2023-12-06
AWS: change how the pause are computed

squioc:
Change the way how and when the connector can go to pause. I propose to use the lag of the collection of events to know if the connector can take a break.


TOUFIKIzakarya (approved):

squioc:
These modifications were applied in another PR. I close this one

PR #621, opened 2023-12-06
Add action to add an IOC to an IOC collection

TonioRyo:
Action to add an IOC to an existing IOC collection. 

Regarding the `ioc_type`, I think it would be best to have the user specify the type of indicator in the action instead of using the `one_per_line` format to try to limit the mistakes of adding improper values. I wanted to provide a drop-down list here but I don't think it's supported in the platform.

Another approach could have been to an action per IOC type but I don't think it would be practical. 

@Darkheir @squioc feel free to share your thoughts and improvement ideas on this, it's a first draft to start the discussion :)

### DIFF
--- a/Sekoia.io/tests/test_add_ioc_2_ioc_collection.py
+++ b/Sekoia.io/tests/test_add_ioc_2_ioc_collection.py
@@ -22,42 +22,33 @@ def arguments_failure():
     }
 
 
-@pytest.fixture
-def add_ioc_response():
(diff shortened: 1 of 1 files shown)


Darkheir (commented):
Version in manifest also needs to be bumped

Darkheir (approved):
Code looks good to me, bt you will need to add unit tests to pass the code coverage gate

PR #627, opened 2023-12-13
Add Netskope Transaction Events connector

lvoloshyn-sekoia:


Netskope/netskope_modules/connector_pubsub_lite.py
```diff
@@ -0,0 +1,144 @@
...
+class PubSubLiteConfig(BaseModel):
+    project_id: str
+    cloud_region: str
+    zone_id: str | None = None
+    subscription_id: str
+
+    frequency: int = 20
+    intake_server: str = "https://intake.sekoia.io"
+    intake_key: str
+    chunk_size: int = 1000
```
squioc, line 23: I suggest importing `DefaultConnectorConfiguration` from `sekoia_automation.connector` and use it:
```suggestion
class PubSubLiteConfig(DefaultConnectorConfiguration):
    project_id: str
    cloud_region: str
    zone_id: str | None = None
    subscription_id: str

    frequency: int = 20
    chunk_size: int = 1000
```

squioc (changes requested):
Nice job.
Some improvements suggested.

squioc (commented):
As the forwarding of events is separated from the collect, I fear that the case of the shutdown of the connector is not well handle in the forwarding task.

squioc (approved):
LTGM

PR #628, opened 2023-12-14
Get the real userAccountControl

TOUFIKIzakarya:
- Improve the way to disable and enable account by getting the real userAccountControl. (We used just 514 and 512 numbers)
- Add an Exception if there's no permission to do some actions


squioc (approved):
LGTM

PR #635, opened 2023-12-14
Jsonify records

TOUFIKIzakarya:
- Replace dict by json records

### DIFF
--- a/AWS/connectors/s3/trigger_s3_cloudfront.py
+++ b/AWS/connectors/s3/trigger_s3_cloudfront.py
@@ -3,7 +3,7 @@
 from connectors.s3 import AbstractAwsS3QueuedConnector, AwsS3QueuedConfiguration
 from itertools import groupby, islice
 
-from typing import Any, List, Dict, Union
+from typing import Any, List, Dict
 import pandas as pd
 import time
 import datetime


AWS/connectors/s3/trigger_s3_cloudfront.py
```diff
@@ -7,6 +7,7 @@
 import pandas as pd
 import time
 import datetime
+import json
```
squioc, line 10: For performance reason, I suggest using `orjson`:
```suggestion
import orjson
```

AWS/connectors/s3/trigger_s3_cloudfront.py
```diff
@@ -67,11 +68,11 @@ def add_start_end_time(self, result: List[Dict[str, Any]], agg_time: float) -> N
             item.update({"end_time": end_time})
             del item["time"]
 
-    def records_to_str(self, results: List[Dict[str, Any]]) -> List[str]:
+    def records_to_json(self, results: List[Dict[str, Any]]) -> List[str]:
         """
         Transform the records to str data.
         """
-        return [str(result) for result in results]
+        return [json.dumps(result) for result in results]
```
squioc, line 75:
```suggestion
        return [orjson.dumps(result) for result in results]
```

squioc (approved):

PR #641, opened 2023-12-18
Hotfix: VadeSecure las_message_id

vg-svitla:


squioc (changes requested):
Can you bump the version number in the manifest (and update the CHANGELOG)? Thanks

squioc (approved):

PR #642, opened 2023-12-19
Fix incorrect parameter name in Add IOC to IOC collection manifest

TonioRyo:


Darkheir (commented):
You should also bump the version in the manifest

PR #646, opened 2023-12-19
fix: Rely on posixpath instead of urllib.parse to join urls

Darkheir:
Replaces `urllib.parse.urljoin` by `posixpath.join` so it work with sub folder.
```py
>>> from urllib.parse import urljoin
>>> urljoin("https://fra2.sekoia.io/api/", "v2/inthreat")
'https://fra2.sekoia.io/api/v2/inthreat'
>>> from posixpath import join as urljoin
>>> urljoin("https://fra2.sekoia.io/api/", "v2/inthreat") 
'https://fra2.sekoia.io/api/v2/inthreat'
```

@squioc do you think we may have places where `urllib.parse.urljoin` was used to remove the previous path ?

### DIFF
--- a/Checkpoint/client/http_client.py
+++ b/Checkpoint/client/http_client.py
@@ -1,7 +1,7 @@
 """Contains http client."""
 from collections.abc import AsyncGenerator
 from datetime import datetime, timezone
-from urllib.parse import urljoin
+from posixpath import join as urljoin
 
 from aiolimiter import AsyncLimiter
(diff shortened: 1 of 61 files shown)


squioc (approved):
@Darkheir No, the method was used to always join the url of the customer tenant and the endpoints of the API.

goudyj:
Hello @squioc, could we push this fix in production soon ?

PR #650, opened 2023-12-21
Fix: Update module to use more logs

vg-svitla:


squioc (changes requested):
Can you bump the version number in the manifest?

squioc (approved):

PR #658, opened 2024-01-04
Sekoia.io: Read v2 notifications from `liveapi`

otetard:
Update the Sekoia.io module to handle “v2 notifications”. This required to change a few things in the existing codebase.

Changes:
- In `_SEKOIANotificationBaseTrigger`, move away from the `handle_${event_type}` and only rely on a `handle_event` function that should be in charge of filtering the collected event. This way, the `SecurityAlertsTrigger` can handle alert notifications (`alert:created`, `alert:updated`, and `alert-comment:created` kind of events).
- Update `AlertCommentCreatedTrigger` to only handle changes about alert status changed (status changes are now in `alert:updated` events).
- Add a new `on_open()` callback to ask `liveapi` to send v2 notifications.
- Update test samples to be v2 notifications.

### DIFF
--- a/Sekoia.io/sekoiaio/triggers/alerts.py
+++ b/Sekoia.io/sekoiaio/triggers/alerts.py
@@ -12,33 +12,37 @@ from .base import _SEKOIANotificationBaseTrigger
 
 class SecurityAlertsTrigger(_SEKOIANotificationBaseTrigger):
     # List of alert types we can handle.
-    HANDLED_EVENT_SUB_TYPES = [
-        "alert-created",
-        "alert-updated",
-        "alert-status-changed",
-        "alert-comment-created",
-    ]
+    HANDLED_EVENT_SUB_TYPES = [("alert", "created"), ("alert", "updated"), ("alert-comment", "created")]
 
-    def handle_alert(self, message):
+    def handle_event(self, message):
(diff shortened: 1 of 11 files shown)


Sekoia.io/sekoiaio/triggers/alerts.py
```diff
@@ -12,33 +12,37 @@
...
+
+        alert_uuid: str | None = None
+        if event_type == "alert":
+            alert_uuid = alert_attrs.get("uuid")
+        elif event_type == "alert-comment":
+            alert_uuid = alert_attrs.get("alert_uuid")
 
         # Is the notification in a format we can understand?
-        if not (alert_event and alert_uuid):
+        if not (event_type and event_action and alert_uuid):
```
Darkheir, line 42: We already checked that `event_type` and `event_action` are in `HANDLED_EVENT_SUB_TYPES` so I guess they can't be None

```suggestion
        if not alert_uuid:
```

Darkheir (approved):

Sekoia.io/sekoiaio/triggers/alerts.py
otetard: Thanks, PR was updated.

PR #659, opened 2024-01-04
Sekoia.io: Explictly log all Alert API exceptions

otetard:
If an error is raised when trying to fetch an alert from the Alert API (after 10 retries), then, log explicitly the error that was raised.

### DIFF
--- a/Sekoia.io/sekoiaio/triggers/alerts.py
+++ b/Sekoia.io/sekoiaio/triggers/alerts.py
@@ -45,7 +45,11 @@ class SecurityAlertsTrigger(_SEKOIANotificationBaseTrigger):
         if not self._filter_notifications(message):
             return
 
-        alert = self._retrieve_alert_from_alertapi(alert_uuid)
+        try:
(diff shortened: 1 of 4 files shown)


Sekoia.io/sekoiaio/triggers/alerts.py
```diff
@@ -45,7 +45,12 @@ def handle_event(self, message):
         if not self._filter_notifications(message):
             return
 
-        alert = self._retrieve_alert_from_alertapi(alert_uuid)
+        try:
+            alert = self._retrieve_alert_from_alertapi(alert_uuid)
+        except Exception as exp:
+            self.log(message="Failed to get fetch alert from the Alert API", level="error")
```
Darkheir, line 51:
```suggestion
```

Darkheir (approved):

Sekoia.io/sekoiaio/triggers/alerts.py
otetard: Fixed.

PR #668, opened 2024-01-09
Add Office365 log collector module

a-leonardi:
It is adapted from the one in SicConfAPI. The connector uses the
Microsoft API to fetch content and forward them to Sekoia.
The existing Office365 module has been split in two submodules to
separate the already-existing one called message_trace and the new one
called management_api.
As a difference with the old SicConfAPI version, no event caching is
done since we cache the last pull date and expect the API to not
duplicate events.

Parent task: [#45065](https://github.com/SekoiaLab/platform/issues/45065)


.pre-commit-config.yaml
otetard, line 1: This file is not related to this PR, could be removed.

otetard (approved):

PR #746, opened 2024-01-19
CrowdStrikeFalcon: Fix stop of the collect when facing ratelimit

squioc:
After a ratelimit on the authentication, the connector stopped collecting the events because `f_stop` was set at the end of the `run` method.

Changes applied:
- change the way to manage thread:
   - change the way to handle graceful shutdown
   - outsource the reading of the events queue in a dedicated and stoppable thread
- remove the expression that set `f_stop` at the end of the `run` method
- Add a timer to sleep 1 minute if the connector faces a ratelimit.

### DIFF
--- a/CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
+++ b/CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
@@ -558,5 +558,11 @@ class EventStreamTrigger(Connector):
                 self.stop_streams(streams, stream_threads)
                 read_queue_thread.stop()
 
+        except HTTPError as error:
+            if error.response is not None and error.response.status_code == 429:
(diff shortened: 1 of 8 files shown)


CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
```diff
@@ -201,7 +211,7 @@ def run(self) -> None:
 
         next_refresh_at = datetime.utcnow() + timedelta(seconds=self.refresh_interval)
         self.log(
-            message=f"reading event stream {self.data_feed_url} starting on offset {self.offset}",
+            message=f"Readin event stream {self.data_feed_url} starting at offset {self.offset}",
```
Darkheir, line 214:
```suggestion
            message=f"Reading event stream {self.data_feed_url} starting at offset {self.offset}",
```

Darkheir (approved):

PR #755, opened 2024-01-25
add connector manifests and bump version of module manifest

ysindel:


Darkheir (changes requested):
All the triggers are not connectors.

For exemple, IPtoASN is a "real" trigger, and the same goes for Sekoia.io's triggers, RSS, MWDB, ....

Darkheir (approved):

PR #776, opened 2024-01-31
get_ioc_collection

rombernier:


rombernier:
Hello @squioc, can you check this PR, also black doesn't pass on a file that I didn't modify can you check would reformat /home/runner/work/automation-library/automation-library/Sekoia.io/sekoiaio/intelligence_center/actions.py
 I can fix this file with black on local ? Thanks

PR #865, opened 2024-02-12
Fix/falcon timer

TOUFIKIzakarya:
Add timer thread to from the SDK to event reader


CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
```diff
@@ -161,8 +162,14 @@ def __init__(
         self.app_id = app_id
         self._stop_event = threading.Event()
         self.events_queue = connector.events_queue
+        self.refresh_timer = RepeatedTimer(self.refresh_interval, self.refresh_stream(refresh_url=self.stream_info["refreshActiveSessionURL"]))  # type: ignore
```
squioc, line 165: The return of `self.refresh_stream(refresh_url=self.stream_info["refreshActiveSessionURL"])` will not be a callable, as expected by the RepeatedTimer, but the result of the call of the `refresh_stream` method (that is None). So this line will result in `self.refresh_timer = RepeatedTimer(self.refresh_interval, None)` (and the refresh stream method will be called only once, at the initialization of the class).

You have to define a new method that call `self.refresh_stream(refresh_url=self.stream_info["refreshActiveSessionURL"])` and use this method in the RepeatedTime (i.e `self.refresh_timer = RepeatedTimer(self.refresh_interval, self.my_refresh_method)`)

squioc (changes requested):
There is an issue in the way to initialize the repeated timer.

CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
TOUFIKIzakarya: Thanks for the remark Sébastien.

squioc (approved):

PR #866, opened 2024-02-12
Add ExtraHop module

lvoloshyn-sekoia:


ExtraHop/extrahop/reveal_360_trigger.py
```diff
@@ -0,0 +1,156 @@
...
+        offset = 0
+
+        while self.running:
+            events = self.fetch_page(from_time=from_time, offset=offset)
+            if not events:
+                self.log(
+                    message=f"No events to forward. Waiting {self.configuration.frequency} seconds",
+                    level="info",
+                )
+                time.sleep(self.configuration.frequency)
```
squioc, line 96: Do you think we can pause the connector in the `next_batch` method? I fear pausing the connector here, after fetching the last page, will prevent us to catch back in we have some delay.

ExtraHop/trigger_extrahop_reveal_360.json
```diff
@@ -0,0 +1,38 @@
+{
+  "description": "Get last logs from the ExtraHop Reveal(x) 360",
+  "docker_parameters": "extrahop_reveal_360",
+  "name": "Fetch new alerts from ExtraHop Reveal(x) 360",
```
squioc, line 4:
```suggestion
  "name": "[BETA] Fetch new alerts from ExtraHop Reveal(x) 360",
```

squioc (changes requested):
Thanks for the jobs.
I proposed some changes and made some suggestions.

ExtraHop/trigger_extrahop_reveal_360.json
squioc, line 1: One last thing. Can you:
 - duplicate this file
 - change the prefix from `trigger_` to `connector_`  in the name of the duplicate
 - remove the property `results` in the duplicate
 ?
 This change is part of a transition on connectors (e.g.: [DUO](https://github.com/SEKOIA-IO/automation-library/tree/main/Duo))

squioc (changes requested):

squioc (approved):
LGTM

PR #868, opened 2024-02-15
feat(AWS): allowing ec2 events with custom unsupported events

r1chev:
Proposal to add custom filtering for ec2 events instead of the default one.

For detection purpose we need most of the Describe* events from ec2.
We defined an unsupported list of events we do not need instead and are quite huge currently in terms of EPS.


squioc (approved):
LGTM

PR #875, opened 2024-02-21
Add exceptions to http response

TOUFIKIzakarya:
Add some logs to see in which line we had the problem and also to avoid printing the exception in the logs section.

### DIFF
--- a/CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
+++ b/CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
@@ -242,31 +242,41 @@ class EventStreamReader(threading.Thread):
                     raise StreamNotAvailable(http_response)
 
                 while self.running:
-                    for line in http_response.iter_lines():
-                        if line.strip():
-                            try:
-                                decoded_line = line.strip().decode()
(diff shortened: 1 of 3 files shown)


CrowdStrikeFalcon/crowdstrike_falcon/event_stream_trigger.py
```diff
@@ -242,31 +242,41 @@ def run(self) -> None:
...
+                            # we exit the loop if the worker is stopping
+                            if not self.running:
+                                break
+                    except Exception as any_exception:
+                        logger.error(
+                            "failed to read line from event stream",
+                            line=line,
+                            stream_root_url=self.stream_root_url,
+                        )
+                        break
```
squioc, line 279: I suggest capturing only the StreamError and lowering the level of the message:
```suggestion
                    except StreamConsumedError:
                        logger.warn(
                            "The datafeed was closed. Reopen it",
                            stream_root_url=self.stream_root_url,
                        )
                        break
```
Could you add the following import also:
```
from requests.exceptions import StreamConsumedError
```

squioc (changes requested):
I suggest a change about what to capture and how to show the error.

PR #881, opened 2024-02-29
Lacework - Fix on timezones

LenaigKaliou:


squioc (changes requested):
Can you update the version number in the manifest file and add an entry in the CHANGELOG.md? Thanks

squioc (approved):
LGTM

PR #883, opened 2024-03-01
Add fastly waf audit connector

lvoloshyn-sekoia:


squioc (commented):
The audit logs from the corp and from the sites will cover distinct parts.
We should get them all.

Fastly/fastly/connector_fastly_waf_audit.py
```diff
@@ -0,0 +1,32 @@
...
+            return (
+                f"{self.base_uri}/api/v0/corps/{self.configuration.corp}/sites/{self.configuration.site}/activity?"
+                f"sort=asc&limit={self.configuration.chunk_size}&from={from_timestamp}"
+            )
+
+        else:
+            return (
+                f"{self.base_uri}/api/v0/corps/{self.configuration.corp}/activity?"
+                f"sort=asc&limit={self.configuration.chunk_size}&from={from_timestamp}"
+            )
```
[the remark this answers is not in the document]
lvoloshyn-sekoia: Corpos and sites [have different types of events](https://docs.fastly.com/en/ngwaf/monitoring-account-activity-with-audit-logs#corp-audit-log), I'm not sure it make sense to mix them altogether. But, perhaps, it would be useful to get events from multiple sites in one connector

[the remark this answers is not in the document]
Adamowoc: We already had different kind of events retrieved from the same connector (ex: Darktrace).

Fastly/fastly/connector_fastly_waf_base.py
```diff
@@ -0,0 +1,305 @@
...
+            )
+
+        # If site is provided - use it
+        # If not - get all sites in the corp
+        if self.configuration.site is not None:
+            sites = [self.configuration.site]
+
+        else:
+            sites = self.get_sites()
+            print("sites", sites)
```
squioc, line 248: I suggest removing this line or maybe define a log that display the count of discovered sites:
```suggestion
```

Fastly/fastly/connector_fastly_waf_base.py
squioc, line 1: Can we use [these helpers](https://github.com/SEKOIA-IO/sekoia-automation-sdk/blob/main/sekoia_automation/connector/workers.py) to manage the Consumers?

squioc (changes requested):
Thanks for the job.
A little fix to apply about a `print` that's lying around.

Fastly/fastly/connector_fastly_waf_base.py
lvoloshyn-sekoia: We can apply them, but I believe `Workers` class will only add more code, because each thread is customized (while `Workers` creates bunch of workers with the same setup). And we'll have to put all of them into a list and go through them to properly supervise, anyway.

squioc (approved):
LGTM

PR #884, opened 2024-03-01
ExtraHop: change the way to compute the url of the endpoint

squioc:
Use `posixpath.join` to compute the url of the endpoint, in order to avoid issue when the customer set the base_url with an ending slash.

### DIFF
--- a/ExtraHop/extrahop/reveal_360_trigger.py
+++ b/ExtraHop/extrahop/reveal_360_trigger.py
@@ -69,7 +69,7 @@ class ExtraHopReveal360Connector(Connector):
             "Content-Type": "application/json",
             "Accept": "application/json",
         }
-        url = urljoin(self.module.configuration.base_url.lstrip("/"), "api/v1/detections/search")
+        url = urljoin(self.module.configuration.base_url.rstrip("/"), "api/v1/detections/search")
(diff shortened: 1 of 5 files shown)


ExtraHop/extrahop/reveal_360_trigger.py
```diff
@@ -68,7 +69,7 @@ def fetch_page(self, from_time: int, offset: int) -> list[dict] | None:
             "Content-Type": "application/json",
             "Accept": "application/json",
         }
-        url = f"{self.module.configuration.base_url}/api/v1/detections/search"
+        url = urljoin(self.module.configuration.base_url.lstrip("/"), "api/v1/detections/search")
```
lvoloshyn-sekoia, line 72: I believe it should be `rstrip`, because we remove a slash in the end of URL

lvoloshyn-sekoia: Now I think that I didn't make this foolproofed enough before, because user can add just a hostname without schema or use an insecure "http://". Perhaps, we could use something like:
```
base_url  = "https://%s" % self.module.configuration.base_url.lstrip("https://").lstrip("http://")
```

lvoloshyn-sekoia (approved):

PR #887, opened 2024-03-05
feat(AWS): filtering some ec2 events with too much EPS

r1chev:
Following previous updates and after several days of telemetry, we are adding some filtering on top EPS event action.


squioc (changes requested):
Can you bump the version number in the manifest file and add an entry in the changelog? Thanks.

squioc (approved):
LGTM

PR #890, opened 2024-03-08
Update package and add exception

TOUFIKIzakarya:
This fix is to handle an request an exception in the google connector.
- Update the library
- Add the exception to the connector


squioc (approved):
LGTM

PR #905, opened 2024-03-26
Improve VadeSecure M365

lvoloshyn-sekoia:


squioc (changes requested):
Thanks for this improvement.
Can you update the version number in the manifest file and add an entry in the Changelog?

squioc (approved):
LGTM

PR #906, opened 2024-03-27
Lacework: Update the module configuration

squioc:
Update the module configuration to match the lacework terminology.

For an APIKey, we get the KeyId, the secret and the account (i.e, the base url of the Lacework tenant). In order to avoid misconfiguration from users, we replace the old parameters names, in the module configuration with these terms.

### DIFF
--- a/Lacework/lacework_module/base.py
+++ b/Lacework/lacework_module/base.py
@@ -4,7 +4,7 @@ from sekoia_automation.module import Module
 
 
 class LaceworkConfiguration(BaseModel):
-    secret: str = Field(secret=True, description="The secret of your API Key")
+    secret: str = Field(..., description="The secret of your API Key", secret=True)
     key_id: str = Field(..., description="The KeyId of your API Key")
     account: str = Field(..., description="The account of your API Key (e.g: `YourLaceworkTenant.lacework.net`)")
 
--- a/Lacework/lacework_module/base.py
+++ b/Lacework/lacework_module/base.py
@@ -4,9 +4,9 @@ from sekoia_automation.module import Module
 
 
 class LaceworkConfiguration(BaseModel):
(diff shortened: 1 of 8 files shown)


Lacework/manifest.json
```diff
@@ -3,32 +3,32 @@
...
-               "secret_key": {
-                   "description": "Client Secret",
+               "secret": {
+                   "description": "The secret of your API Key",
                    "type": "string"
                },
-               "lacework_url": {
-                   "description": "Client Lacework application name",
+               "account": {
+                   "description": "The account of your API Key",
```
lvoloshyn-sekoia, line 15:
```suggestion
                   "description": "The account of your API Key (e.g: `YourLaceworkTenant.lacework.net`)",
```
Perhaps, we should provide example here as well?

Lacework/lacework_module/base.py
```diff
@@ -4,9 +4,9 @@
 
 
 class LaceworkConfiguration(BaseModel):
-    secret_key: str = Field(secret=True, description="Your unique identifier used as the Authorization header")
-    access_key: str = Field(..., description="Your unique key to create the token used to authenticate the API Key.")
-    lacework_url: str = Field(..., description="Your Lacework application name")
+    secret: str = Field(secret=True, description="The secret of your API Key")
```
lvoloshyn-sekoia, line 7:
```suggestion
    secret: str = Field(..., description="The secret of your API Key", secret=True)
```
The field still has to be required (though, idk whether this change will affect anything)

lvoloshyn-sekoia (changes requested):
Thanks, looks good. Few very small notes:

lvoloshyn-sekoia (approved):
LGTM

PR #907, opened 2024-03-29
Fix: azure blob chunk size

vg-svitla:


Azure/connectors/azure_blob.py
```diff
@@ -95,6 +96,7 @@ async def get_azure_blob_data(self) -> list[str]:
         blob_list = self.azure_blob_wrapper().list_blobs()
         _last_modified_date = self.last_event_date
         records: list[Any] = []
+        limit_of_events_to_push = int(os.getenv("AZURE_BATCH_SIZE", 10000))
```
squioc, line 99: Cannot we move the definition of `limit_of_events_to_push` in `__init__`? In order to avoid having to read the env variable each time.

Azure/connectors/network_watcher.py
```diff
@@ -96,6 +97,7 @@ async def get_azure_network_watcher_data(self) -> list[str]:
         blob_list = self.azure_blob_wrapper().list_blobs()
         _last_modified_date = self.last_event_date
         records: list[Any] = []
+        limit_of_events_to_push = int(os.getenv("AZURE_BATCH_SIZE", 10000))
```
squioc, line 100: Same as above

AWS/connectors/s3/__init__.py
```diff
@@ -102,6 +102,7 @@ async def next_batch(self, previous_processing_end: float | None = None) -> tupl
             tuple[list[str], int]:
         """
         records = []
+        limit_of_events_to_push = os.getenv("AWS_BATCH_SIZE", 10000)
```
squioc: Same than Azure

Azure/connectors/azure_blob.py
vg-svitla: I thought about this
In current workflow we can change env varibale on the fly. Maybe it will be more useful for debug purposes

squioc (approved):
LGTM

PR #912, opened 2024-04-03
Fix Imperva module connector

StephKll3c:


StephKll3c:
> you might need to bump the version ?
Just run black :innocent:

PierrickV (approved):

PR #914, opened 2024-04-09
Fix: Azure Blob exclude empty lines

vg-svitla:


Adamowoc (approved):
Thanks :)

squioc (commented):
Can you add a test?

squioc (approved):
LGTM

PR #917, opened 2024-04-11
Improve timestamp and exceptions

TOUFIKIzakarya:
- Change timestamp storage way
- Add some exceptions
- Add some logs 
- Delete some unucessary packages
- Update google package


squioc (approved):
LGTM

PR #918, opened 2024-04-12
Feature: Azure Key Vault

vg-svitla:


squioc (commented):
LGTM. Thanks for the job.
However, Can we mutualize the code for Azure blob connectors like we did for [AWS S3 connectors](https://github.com/SEKOIA-IO/automation-library/blob/main/AWS/connectors/s3/__init__.py)? i.e having a base class with the whole stuff and just having a method to parse the content of the blob in each  connector?

squioc (approved):
LGTM

PR #926, opened 2024-04-22
fix(withsecure): add read right to oauth + apply ruff

goudyj:


penhouetp (approved):

goudyj:
@squioc could you deploy it please?

PR #943, opened 2024-05-14
Fixed: Add linter to ci and fix all jsons

vg-svitla:


linter.py
```diff
@@ -0,0 +1,86 @@
...
+
+    result = subprocess.run(
+        ["npx", "prettier", action] + files_to_fix,
+        capture_output=True
+    )
+
+    if result.returncode != 0:
+        raise Exception("Prettier failed: \n" + result.stderr.decode())
+    else:
+        print("Prettier run successfully: \n" + result.stdout.decode())
```
squioc, line 56: Can we not rely on the python json module (like [this example](https://github.com/SEKOIA-IO/intake-formats/blob/main/utils/linter.py#L63)) instead of `prettier`? What do you think about? It will avoid contributors to install `prettier`.

vg-svitla: Use json.dumps instead

squioc (approved):
LGTM

PR #947, opened 2024-05-16
GA: add the name of the module in the name of the uploaded artifact

squioc:
add the name of the module in the name of the uploaded artifact to avoid conflict when several modules are updated in a PR (see [GA job](https://github.com/SEKOIA-IO/automation-library/actions/runs/9109729684/?pr=943))

### DIFF
--- a/.github/workflows/build-modules.yml
+++ b/.github/workflows/build-modules.yml
@@ -94,7 +94,7 @@ jobs:
       - name: Upload Event
         uses: actions/upload-artifact@v4
         with:
-          name: Event File
+          name: Event File ${{ matrix.module }}
           path: ${{ github.event_path }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-modules.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       - name: Upload Event
         uses: actions/upload-artifact@v4
         with:
-          name: Event File
+          name: Event File ${{ matrix.module }}
```
otetard, line 97: With this change we will also have to update the action that handles these files:

https://github.com/SEKOIA-IO/automation-library/blob/4bcbe6f474b64c7885f41d41718fcfc58c1ad178/.github/workflows/test_results.yml#L26-L32

PR #948, opened 2024-05-17
Improve get context action

TOUFIKIzakarya:
- Add Output results to the action
- Add a use case to it : When there are many objects on the list


squioc (approved):
LGTM

PR #958, opened 2024-05-28
Connectors: fix configuration

squioc:
Fix the configuration of some connectors for which the intake key parameter was missing


StephKll3c (approved):
Thx :)

PR #961, opened 2024-05-29
Fix: Lacework event date

vg-svitla:


Lacework/lacework_module/lacework_connector.py
```diff
@@ -139,57 +139,55 @@ def get_response_by_timestamp(self) -> Any | None:
...
-            self.most_recent_date_seen = datetime.utcfromtimestamp(last_date)
-            messages = [orjson.dumps(message).decode("utf-8") for message in items]
-            self.log(message=f"Sending the first batch of {len(messages)} messages", level="info")
-            OUTCOMING_EVENTS.labels(intake_key=self.configuration.intake_key).inc(len(messages))
-            self.push_events_to_intakes(events=messages)
+        if first_batch is not None:
+            next_page_url = self.get_next_page_url(first_batch)
+            first_batch_events = first_batch.get("data", [])
+
+            last_date = max([self.parse_event_time(item["endTime"]) for item in first_batch_events])
```
squioc: We still uses `endTime` to save the last date, but we [uses `startTime` when requesting the alerts](https://github.com/SEKOIA-IO/automation-library/blob/main/Lacework/lacework_module/client/__init__.py#L39).
So, I think that the PR will not fix the issue (see Hypothesis in the ticket).

squioc (approved):
LGTM

squioc (approved):
LGTM

PR #963, opened 2024-05-29
Actions for rules

LenaigKaliou:


squioc (changes requested):
Can you add an entry in the CHANGELOG.md and increase the number version in the manifest? Thanks.

squioc (approved):
LGTM

PR #974, opened 2024-06-04
Add Mimecast module

lvoloshyn-sekoia:


Mimecast/mimecast_modules/connector_mimecast_siem.py
```diff
@@ -0,0 +1,249 @@
...
+                self.log(message=f"Restarting `{consumer_name}` consumer", level="info")  # pragma: no cover
+
+                consumers[consumer_name] = MimecastSIEMWorker(connector=self, log_type=consumer_name)
+                consumers[consumer_name].start()
+
+    def stop_consumers(self, consumers: dict[str, MimecastSIEMWorker]):
+        for consumer_name, consumer in consumers.items():
+            if consumer is not None and consumer.is_alive():
+                self.log(message=f"Stopping `{consumer_name}` consumer", level="info")  # pragma: no cover
+                consumer.stop()  # pragma: no cover
```
squioc, line 241: Can we use [workers](https://github.com/SEKOIA-IO/sekoia-automation-sdk/blob/main/sekoia_automation/connector/workers.py) instead of rewriting these methods?
For the `log_type`, we can use [`partial`](https://docs.python.org/3/library/functools.html#functools.partial) to set the `log_type` parameter for the `MimecastSIEMWorker` initiator (that will derivate the class with the `log_type` set).

squioc (approved):
LGTM

PR #980, opened 2024-06-11
Fix: Github connector add paging and refresh token request

vg-svitla:


Github/github_modules/audit_log_trigger.py
```diff
@@ -183,7 +183,8 @@ async def next_batch(self) -> None:
         delta_sleep = self.configuration.frequency - batch_duration
         if delta_sleep > 0:
             logger.debug(f"Next batch in the future. Waiting {delta_sleep} seconds")
```
squioc, line 185: I think we can remove this line as it's a duplication of the following one

squioc (approved):
LGTM

PR #981, opened 2024-06-11
Fix compliance check issues

lvoloshyn-sekoia:
Few notes:
1. To avoid false positive detections on square logos without intended transparent background, such logos are manually edited to have 1 pixel with 254 alpha out of 255 (e.g.  [CrowdStrike Falcon logo ](https://github.com/SEKOIA-IO/automation-library/pull/981/files#diff-d5ae504cc5a8ef6a877115a411307eef484104c5802a9e50040dadc8afe9e0e8))
2. Some modules have updated versions in `manifest.json` (e.g. `1.0.0` instead of `1.0`)


squioc (approved):
LGTM

PR #991, opened 2024-06-17
Automation Compliance tool improvement

lvoloshyn-sekoia:


_utils/compliance/validators/main.py
squioc, line 1: The validator doesn't check if two classes are registered with the same docker_parameter in the same main.py.
Can you add this verification? Thanks.

squioc (approved):
LGTM

PR #992, opened 2024-06-17
Fix compliance issues

lvoloshyn-sekoia:
Fix stuff found by https://github.com/SEKOIA-IO/automation-library/pull/991


Sekoia.io/trigger_sekoiaio_alert_webhook.json
squioc, line 1: We must keep this trigger's manifest (even if there is no python code associated to it in the module).
Maybe, for the compliance check script, we should be able to add a dot-ignore file in the module to exclude some file from the check.

squioc (changes requested):
Can you remove changes in the Sekoia.io automation module? Thanks.

squioc (approved):
LGTM

PR #995, opened 2024-06-19
Nybble / fixing optional field in events but listed in rule

GabrielKropp:
sometimes, events are not containing all fields present in the rule which has triggered the alert.

Fixing the code to handle this.


squioc (commented):
The code looks good to me.
Can you add a litlle entry in the changelog? Thanks.

GabrielKropp:
> The code looks good to me. Can you add a litlle entry in the changelog? Thanks.

Thanks ! changelog entry added

squioc (approved):
LGTM

PR #1008, opened 2024-07-01
Change days to hours in the timestepper

TOUFIKIzakarya:
- Change days to hours in timestepper because we got a big delay.


squioc (approved):
LGTM

PR #1011, opened 2024-07-02
Catch some network exceptions

TOUFIKIzakarya:
- Catch some Network exception :

1. BrokenPipeError: [Errno 32] Broken pipe
2. OSError: [Errno 101] Network is unreachable


squioc (approved):
LGTM

PR #1012, opened 2024-07-02
feat(sekoiaio): add getters for community, entity and intake

gaelmuller:
Create actions to resolve:

* A community
* An entity
* An intake

### DIFF
--- a/Sekoia.io/sekoiaio/workspace/__init__.py
+++ b/Sekoia.io/sekoiaio/workspace/__init__.py
@@ -1,0 +1,14 @@
+from sekoia_automation.action import GenericAPIAction
+
+base_url = "api/v1/"
+
+
+GetCommunity = type(
+    "GetCommunity",
+    (GenericAPIAction,),
+    {
+        "verb": "get",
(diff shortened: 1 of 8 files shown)


Sekoia.io/sekoiaio/workspace/__init__.py
```diff
@@ -0,0 +1,14 @@
+from sekoia_automation.action import GenericAPIAction
+
+base_url = "api/v1/"
+
+
+GetCommunity = type(
```
Darkheir, line 6: Out of curiosity what's the advantage of declaring the class this way instead of doing

```py
class GetCommunity(GenericAPIAction):
    verb = "get"
    endpoint =  base_url + "communities/{uuid}"
    query_parameters = []
```

Darkheir (approved):

Sekoia.io/sekoiaio/workspace/__init__.py
gaelmuller: No idea, done out of consistency here :p

PR #1018, opened 2024-07-04
Fix: Bitsight use max delay for 7 days

vg-svitla:


squioc (changes requested):
Code looks good to me.
Can you update the manifest and the CHANGELOG? Thank you

squioc (approved):
LGTM

PR #1020, opened 2024-07-08
Connectors: fix the way to compute lag

squioc:
- For Darktrace and BitSight, handle the case when no events are collected to compute the event lags.
- For ExtraHop and PaloAlto Cortex XDR, change the precision of the timestamps.


TOUFIKIzakarya (approved):
LGFM

PR #1024, opened 2024-07-09
Fix: Sophos updates

vg-svitla:


squioc (approved):
Thanks for the changes.

PR #1045, opened 2024-07-23
fix(Sekoia.io): Fixes potential errors in URLs

Darkheir:
* Handle cases where the base URL from the config has the `/api` or not.
* Harmonize all the triggers
  * `/me` and `/v2/inthreat` were expecting `/api` to be in the base url (or that the base url is `https://api.sekoia.io`)
  *  `/v1/sic/alerts` was not expecting `/api` to be in the URL

The fix should also work with `https://api.sekoia.io` because `https://api.sekoia.io/api` also works.


squioc (approved):
LGTM

PR #1054, opened 2024-07-30
fix(retarus-stopping): fix connector when encountering error

goudyj:


Retarus/retarus_modules/consumer.py
```diff
@@ -25,23 +30,19 @@ def stop(self):
         """Sets the stop event"""
         self._stop_event.set()
 
-    def create_websocket(self) -> Tuple[WebSocketApp, Thread]:
+    def create_websocket(self) -> WebSocketApp:
         """Creates a WebSocket inside a Thread
 
         Returns:
             Tuple[WebSocketApp, Thread]: The websocket we opened, The thread in which it will run
```
julienCu, line 37: Comment to update

goudyj: Thanks @julienCu indeed !

Retarus/retarus_modules/consumer.py
```diff
@@ -76,13 +77,18 @@ def on_close(self, *_):
     def run(self):
         """Start the websocket thread then wait for the stop event to be set and close the websocket when it happens"""
         self.log(message=f"Connection to stream {self.configuration.ws_url}", level="info")
+        while self.is_running:
+            self.websocket = self.create_websocket()
```
Darkheir, line 81: The websocket seems to be used only here

```suggestion
            websocket = self.create_websocket()
```

Darkheir (commented):
I guess the version in the manifest must be bumped.

Also I'm not sure to understand how the stop of the websocket will work. The sto penvent will be set, but the websocket will be inside the `run_forever` loop.

Darkheir (commented):
I still don't see the version bump in the `manifest.json` file

PR #1059, opened 2024-08-01
Azure & AWS: add two new metrics to monitor events lag

squioc:
- add a metric to monitor the age of the older message seen in a batch
- add a metric to monitor the average age of messages seen in a batch


TOUFIKIzakarya (approved):
LGFM

PR #1060, opened 2024-08-02
Remove need for a configuration for Sekoia modules

a-leonardi:
To achieve internal/automatic configuration of the Sekoia.io modules, they have to *not* require a configuration from the user.
Their configuration elements have been removed.
This configuration will be created and linked to the module in Symphony API but won't be accessible to the user.
Only the `secrets` field has been kept in order to encrypt the relevant field when it will be automatically created.

Parent task: [#58166](https://github.com/SekoiaLab/platform/issues/58166)


Darkheir (approved):
Code looks good, but let's wait that the PR inside the API is merged before merging this one :-)

PR #1061, opened 2024-08-02
Jsonify events

TOUFIKIzakarya:
Issue : [#160](https://github.com/SekoiaLab/integration/issues/160#event-13739491508)

### DIFF
--- a/Sophos/sophos_module/trigger_sophos_xdr_query.py
+++ b/Sophos/sophos_module/trigger_sophos_xdr_query.py
@@ -192,7 +192,6 @@ class SophosXDRQueryTrigger(SophosConnector):
                 self.events_sum += len(messages)
 
                 while next_key:
-                    grouped_data = []
                     response_next_page: dict[str, Any] = self.client.get_query_results(
                         query_id, pagination, next_key
                     ).json()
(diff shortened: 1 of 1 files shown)


Sophos/sophos_module/trigger_sophos_xdr_query.py
```diff
@@ -198,11 +198,12 @@ def getting_results(self, pagination: int) -> None:
...
 
                     next_page_items = response_next_page.get("items", [])
+                    next_messages = [orjson.dumps(message).decode("utf-8") for message in next_page_items]
                     self._observe_items_events_lag(next_page_items)
-                    grouped_data.extend(next_page_items)
+                    grouped_data.extend(next_messages)
                     self.log(message=f"Sending other batches of {len(grouped_data)} messages", level="info")
                     OUTCOMING_EVENTS.labels(intake_key=self.configuration.intake_key).inc(len(grouped_data))
-                    self.push_events_to_intakes(events=grouped_data)
+                    self.push_events_to_intakes(events=next_messages)
```
squioc, line 206: It should not stay `grouped_data` here?
You update line 201 in order that the messages pushed in `grouped_data` are JSON document.
Unless `grouped_data` is useless and can be replaced by `next_messages` (it seems to be the case).

```suggestion
                    self.push_events_to_intakes(events=grouped_data)
```

TOUFIKIzakarya: Yes normally it should be `next_messages `. So i will delete `grouped_data`.

squioc (approved):

PR #1062, opened 2024-08-02
Netskope: handle when failing to decode the incoming message

squioc:
Handle the error when the incoming message is not a UTF-8 string.


TOUFIKIzakarya (approved):
LGFM

PR #1072, opened 2024-08-08
SentinelOne: filter collected events

squioc:
Filter the events in order to discard already collected ones. We use a LRU cache to save temporary the identifier of the last colllected events.


TOUFIKIzakarya (approved):
LGFM

PR #1075, opened 2024-08-12
feat(Sekoia.io): Enforce default and limit from manifests

Darkheir:
A default limit was not applied on the first query when it was not provided in the arguments. 
Also the max in the code was set to `1 000` and in the manifest it was `100`. I updated the code to reflect the manifest, but we can pass it to `1 000` if you feel `100` is too low

This PR also raise the default limit to `100`.


squioc (approved):
LGTM

PR #1077, opened 2024-08-21
Office365 - log more details for auth errors

lvoloshyn-sekoia:
To fix https://github.com/SekoiaLab/integration/issues/184 we have to get more details


TOUFIKIzakarya (commented):
Can you please update the manifest file and also the changelog

TOUFIKIzakarya (approved):
LGFM

PR #1078, opened 2024-08-23
WithSecure - add actions

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/178


WithSecure/withsecure/response_action.py
```diff
@@ -0,0 +1,50 @@
...
+            stop_event=Event(),
+            log_cb=self.log,
+        )
+        response = client.post(
+            API_RESPONSE_ACTIONS_URL,
+            timeout=API_TIMEOUT,
+            json=request_payload,
+            headers=headers,
+        )
+        response.raise_for_status()
```
squioc, line 48: I think we should extract the message in the failed response (maybe in the `run` method):
```json
{
  "message": "Bad request because of invalid parameters or data.",
  "code": 400000,
  "transactionId": "0000-d6e9bdb9f82946f3"
}
```

squioc (changes requested):
Thanks for the contribution.
Some changes to apply

WithSecure/withsecure/response_action.py
```diff
@@ -0,0 +1,50 @@
...
+    ) -> Any:
+        self.log(
+            f"Execute the action type '{action_type}' on device '{target}'",
+            level="debug",
+        )
+
+        request_payload: dict[str, Any] = {
+            "type": action_type,
+            "organizationId": organization_id,
+            "targets": [target],
```
squioc, line 25: According to some tests I made with the API, the field `parameters` is required (even empty)
```suggestion
            "targets": [target],
            "parameters": {},
```

squioc (changes requested):

squioc (approved):
LGTM

PR #1079, opened 2024-08-26
SentinelOne: fix connector

squioc:
- Handle threats as a dict (and not an object with properties).
- declare `batch_duration` as we use it to compute the pause.


TOUFIKIzakarya (approved):
LGFM

PR #1080, opened 2024-08-26
CrowdStrikeFalcon: prevent events collect failure when the verticles collector authentication fails

squioc:


TOUFIKIzakarya (approved):
Nice one 👌🏻

PR #1087, opened 2024-09-09
Feature: Add support of isolate and deisolate host in Crowdstrike Falcon

vg-svitla:


squioc (approved):
Thanks for the work.
LTGM

PR #1089, opened 2024-09-10
feat(Triage) option to exclude signed binary

r1chev:
First attempt to add an option in Triage trigger to give the possibility to exclude some reports based on signed binary. 

Most of the time when a signed binary is used as entry point in a sandbox submission, it results in false positives in the extracted IOCs.

By checking the static report of the related submission we could check if there is a `trusted` signature in it. In that case no IOCs will be return.


Triage/manifest.json
```diff
@@ -15,6 +15,10 @@
         "type": "string",
         "description": "API endpoint (url)",
         "format": "uri"
+      },
+      "exclude_signed": {
+        "type": "boolean",
+        "description": "Do not retrieve report with signed binary"
       }
```
Darkheir, line 22: Shouldn't it be a configuration for the trigger and not the account ?

Also we may want to provide a default value to not break the existing ones

r1chev: You're right thx!

Darkheir: It should be moved inside `trigger_triage.json`

Triage/triage_modules/trigger_triage.py
```diff
@@ -37,6 +37,10 @@ def api_key(self):
     def api_url(self):
         return self.module.configuration["api_url"]
 
+    @property
+    def exclude_signed(self):
+        return self.configuration.get("exclude_signed", False)
```
squioc, line 42: According to this line, It confirms that the parameter must move to the trigger configuration

squioc (changes requested):
Thanks for the contribution.
Can you add an entry to the changelog also? Thanks.

squioc (approved):
LGTM

Darkheir (approved):

PR #1095, opened 2024-09-18
Fix: Trellix change token refresh and loop init

vg-svitla:


Trellix/connectors/trellix_edr_connector.py
```diff
@@ -291,10 +291,11 @@ async def async_run(self) -> None:  # pragma: no cover
 
     def run(self) -> None:  # pragma: no cover
         """Runs TrellixEdr."""
+        loop = asyncio.get_event_loop()
+
         while self.running:
             try:
-                loop = asyncio.get_event_loop()
                 loop.run_until_complete(self.async_run())
```
squioc, line 298: If an exception happens in `async_run`, the loop won't be closed?
I remember that we did [some double-loops](https://github.com/SEKOIA-IO/automation-library/blob/main/BroadcomCloudSwg/connectors/broadcom_cloud_swg_connector.py#L520-L524) in some connector to recreate a loop if we face an exception.

squioc (approved):
LGTM

PR #1101, opened 2024-09-20
1Password - fix cold start

lvoloshyn-sekoia:


1Password/manifest.json
```diff
@@ -25,5 +25,5 @@
   "name": "1Password",
   "uuid": "56f9e1f6-95ba-45ed-867b-44fb3183934d",
   "slug": "one-password",
-  "version": "0.1.0"
+  "version": "0.1.1"
```
squioc, line 28:
```suggestion
  "version": "1.0.0"
```

squioc (changes requested):
Can you also add an entry in the changelog? Thanks

squioc (changes requested):
I propose to bump the version number to 1.0.0, as the connector is in production.

squioc (approved):
LGTM

PR #1104, opened 2024-09-24
Fix secret fields in modules

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/209

`Ubika` already has secrets.
`Google` has a convoluted config json, will have to make sure that a new configuration will work.


Google/manifest.json
```diff
@@ -19,6 +19,9 @@
           "auth_provider_x509_cert_url",
           "client_x509_cert_url"
         ],
+        "secrets": [
+          "private_key"
```
squioc, line 23: Shouldn't we add `private_key_id` also?

squioc (approved):
LGTM

PR #1111, opened 2024-09-25
Change the result type

TOUFIKIzakarya:
- Change the type of result to dict due to validation function that support just dict type in automation sdk


squioc (approved):
LGTM

PR #1112, opened 2024-09-26
Azure EventHub - fix epoch error

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/232


squioc (changes requested):
Thanks for the work.
Can you bump the version number and add an entry in the CHANGELOG.md? Thank you

PR #1133, opened 2024-10-17
chore: Use status title instead of uuid when changing alert status

ysindel:
[Parent Task](https://github.com/SekoiaLab/platform/issues/60784)

### DIFF
--- a/Sekoia.io/action_trigger_an_action_on_the_alert_workflow.json
+++ b/Sekoia.io/action_trigger_an_action_on_the_alert_workflow.json
@@ -12,10 +12,17 @@
         "type": "string",
         "in": "path"
       },
-      "action_uuid": {
+      "status": {
         "type": "string",
-        "description": "UUID of the action to trigger",
-        "in": "body"
+        "description": "Title of the status to update the alert to",
+        "in": "body",
+        "enum": [
+          "Pending",
+          "Acknowledge",
+          "Ongoing",
+          "Reject",
+          "Close"
+        ]
       },
       "comment": {
         "type": "string",
--- a/Sekoia.io/action_trigger_an_action_on_the_alert_workflow.json
+++ b/Sekoia.io/action_trigger_an_action_on_the_alert_workflow.json
@@ -18,10 +18,10 @@
         "in": "body",
         "enum": [
           "Pending",
-          "Acknowledge",
+          "Acknowledged",
           "Ongoing",
-          "Reject",
-          "Close"
+          "Rejected",
+          "Closed"
         ]
       },
       "comment": {
--- a/Sekoia.io/sekoiaio/operation_center/update_alert_status.py
+++ b/Sekoia.io/sekoiaio/operation_center/update_alert_status.py
@@ -26,20 +26,25 @@ class UpdateAlertStatus(Action):
 
     @retry(
         reraise=True,
-        wait=wait_exponential(max=10),
-        stop=stop_after_attempt(5),
+        wait=wait_exponential(max=300),
+        stop=stop_after_attempt(10),
     )
     def perform_request(self, alert_uuid: str, status: str):
-
-        result = requests.patch(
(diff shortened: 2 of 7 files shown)


Sekoia.io/sekoiaio/operation_center/update_alert_status.py
```diff
@@ -0,0 +1,45 @@
...
+        result = requests.patch(
+            self.url(alert_uuid), headers=self.headers, json={"action_uuid": STATUS_UUIDS[status.upper()]}
+        )
+        if not result.ok:
+            self.error(f"Could not change alert {alert_uuid} status, status code: {result.status_code}")
+            result.raise_for_status()
+        return result.json()
+
+    def run(self, arguments: dict):
+        status = arguments.get("status")
```
Darkheir, line 43: It would be nice to still support `action_uuid` in the code (not the manifest).

It would avoid to break existing actions configured with `action_uuid`

Sekoia.io/sekoiaio/operation_center/update_alert_status.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def headers(self) -> dict:
+        api_key = self.module.configuration["api_key"]
+        return {"Authorization": f"Bearer {api_key}"}
+
+    @retry(
+        reraise=True,
+        wait=wait_exponential(max=10),
+        stop=stop_after_attempt(5),
```
Darkheir, line 30: Increase time to retry, we already had errors during many minutes. Also maybe we shouln't retry 4xx errors.


```suggestion
        wait=wait_exponential(max=300),
        stop=stop_after_attempt(10),
```

Darkheir (changes requested):
The version in `manifest.json` and the changelog must also be updated.

Sekoia.io/sekoiaio/operation_center/update_alert_status.py
ysindel: Thank you, I'll change these

Sekoia.io/sekoiaio/operation_center/update_alert_status.py
ysindel: oh you're right, should we add a new action to not break existing actions or just support `action_uuid` ?

Darkheir: I think just supporting `action_uuid` will be fine. Having 2 separate actions will confuse the clients.

Once the module has been deploy we can run a migration to update the playbooks using `action_uuid` to rely on `status` instead.

ysindel: Okay I'll restore support for `action_uuid` in the code

Sekoia.io/action_trigger_an_action_on_the_alert_workflow.json
```diff
@@ -12,10 +12,17 @@
...
-      "action_uuid": {
+      "status": {
         "type": "string",
-        "description": "UUID of the action to trigger",
-        "in": "body"
+        "description": "Title of the status to update the alert to",
+        "in": "body",
+        "enum": [
+          "Pending",
+          "Acknowledge",
```
Darkheir, line 21:
```suggestion
          "Acknowledged",
```

Sekoia.io/action_trigger_an_action_on_the_alert_workflow.json
```diff
@@ -12,10 +12,17 @@
...
+          "Ongoing",
+          "Reject",
```
Darkheir, line 23:
```suggestion
          "Rejected",
```

Sekoia.io/action_trigger_an_action_on_the_alert_workflow.json
```diff
@@ -12,10 +12,17 @@
...
+          "Close"
```
Darkheir, line 24:
```suggestion
          "Closed"
```

Darkheir (approved):

PR #1135, opened 2024-10-18
add groupby module

rombernier:
Hello, the goal here is to add a module that performs a 'for each' operation with a 'group by' on a specific key, along with an option to filter by a specific value.
I'm not sure about the use of the yield & the output properties
let me know what you think, thanks


rombernier:
> Personally, I would have added an action_ to the sekoia module rather than creating a new top-level module.

Hello, I created a separate module because I have a small hope of using private integrations to test this out :)

ForEachGroupBy/groupby/action_groupby.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def run(self, arguments):
+        group_key = arguments.get("group_key", 0)  # key to group by
+        filter_key = arguments.get("filter_key", None)  # key to filter by, if provided
+        filter_value = arguments.get("filter_value", None)  # value to filter by, if provided
+        input_data = arguments.get("input", [])  # input array of elements
+
+        # Dictionary to hold grouped elements
+        grouped = {}
+
```
squioc, line 17: Can you check in `input_data` is a list?

squioc (changes requested):
Thanks for the contribution.
I think that the [`Utils`](https://github.com/SEKOIA-IO/automation-library/tree/main/Utils) module is more appropriated than created a new one.
Can you move the action into it? Thanks

ForEachGroupBy/groupby/action_groupby.py
rombernier: Done

rombernier:
Hello @squioc, can you check this PR ? How can I test it on the test platform ? Thanks

penhouetp (approved):

goudyj (approved):

Utils/utils/action_groupby.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # Yield each filtered group with appropriate indexing
+        for group_value, elements in filtered_groups.items():
+            yield {
+                "group_index": group_index,
+                "total_groups": total_groups,
+                "group_value": group_value,  # Added group_value to the output
+                "group_data": elements,
+            }
+            group_index += 1
```
squioc, line 56: I think you must return a list here, not a generator; it isn't supported by the sdk (cf [how results from actions are handled by the sdk](https://github.com/SEKOIA-IO/sekoia-automation-sdk/blob/main/sekoia_automation/action.py#L98-L107)).

Utils/action_groupby.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "description": "GroupBy",
+  "docker_parameters": "action_groupby",
+  "name": "GroupBy",
+  "results": {
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "properties": {},
+    "required": [],
+    "title": "Results",
+    "type": "object"
```
squioc, line 36: According to my comment above
```suggestion
    "type": "array"
```

squioc (changes requested):
A generator returned by the action isn't supported by the sdk.

Utils/utils/action_groupby.py
rombernier: Ok, I will change that

squioc (approved):
LGTM

PR #1136, opened 2024-10-21
Microsoft Secure Graph - add actions

lvoloshyn-sekoia:
1. Adding comment is available in `update alert` endpoint, so I didn't add it as a separate action
2. Added  `cancel action`, because machine actions aren't immediate, they can spend some time in `Pending` state. For instance, when we isolate a machine, we can't unisolate it right away, because `isolate` action is still pending. But we can cancel it.


MicrosoftDefender/microsoftdefender_modules/action_push_indicators.py
```diff
@@ -0,0 +1,138 @@
...
+        url = urljoin(self.client.base_url, "api/indicators/BatchDelete")
+        return self.client.post(url, json={"IndicatorIds": indicators_ids})
+
+    def remove_indicators(self, indicators: list[dict[str, Any]]) -> None:
+        ids_to_remove: list[Any] = []
+
+        for indicator in indicators:
+            q = f"indicatorValue+eq+'{indicator['indicatorValue']}'"
+            response = self.list_indicators(q=q)
+            self.process_response(response)
```
squioc, line 53: I wonder if we can't use the [`in`](https://docs.oasis-open.org/odata/new-in-odata/v4.01/cn04/new-in-odata-v4.01-cn04.html#_Toc21700044) operator?

MicrosoftDefender/microsoftdefender_modules/action_push_indicators.py
```diff
@@ -0,0 +1,138 @@
...
+        response = self.import_indicators(indicators=indicators)
+        self.process_response(response)
+
+        return response
+
+    @staticmethod
+    def get_payload(value: Any, type: Any, args: dict[str, Any]) -> dict[str, Any]:
+        return {
+            "title": "indicator",
+            "createdByDisplayName": "Sekoia.io",
```
squioc, line 72: According to [this table](https://learn.microsoft.com/en-us/defender-endpoint/api/ti-indicator#properties), it seems that this property should be `createdBySource`.
```suggestion
            "createdBySource": "Sekoia.io",
```

squioc (commented):
Thank you for the job.
Some suggestions and comments!!

MicrosoftDefender/microsoftdefender_modules/action_push_indicators.py
lvoloshyn-sekoia: I wasn't able to set `createdBySource` to `Sekoia.io`, got an error. `PublicApi` worked, though

squioc (changes requested):

MicrosoftDefender/microsoftdefender_modules/action_stop_and_quarantine_file.py
squioc, line 1: I was unable to validate this action. The API always answer an internal error:
```
{'error': {'code': 'InternalServerError', 'message': 'Internal Server Error', 'target': '|<REDACTED>.'}}
```
I suggest removing this action from the PR

MicrosoftDefender/action_stop_and_quarantine_file.json
squioc, line 1: As I suggest removing the code, I suggest removing this manifest.

squioc (changes requested):

squioc (approved):
LGTM. Thank you for the work

PR #1138, opened 2024-10-21
standardize module name

pbivic:


goudyj:
You need to update the version as well in manifest

goudyj (approved):

PR #1139, opened 2024-10-22
Fix: Tehtris remove duplicates from result (257)

vg-svitla:
Closes [257](https://github.com/SekoiaLab/integration/issues/257)


Tehtris/tehtris_modules/trigger_tehtris_events.py
```diff
@@ -77,15 +80,34 @@ def __fetch_next_events(self, from_date: datetime, offset: int):
...
+        Remove duplicates events from the fetched events and update the cache with new ids.
+
+        Args:
+            events: list[dict[str, Any]]
+
+        Returns:
+            list[dict[str, Any]]:
+        """
+        result = [event for event in events if event["uid"] not in self.events_cache]
+        self.events_cache.clear()
```
squioc, line 94: Do we need to clear the cache? As an LRU cache, it should manage the key itself.

squioc (commented):
Thanks for the work.

squioc (approved):
LGTM

PR #1142, opened 2024-10-24
Add some other services

TOUFIKIzakarya:
Add some services to config file, ( We support them in the connector but we didn't mention them in config files )
[Issue](https://github.com/SekoiaLab/integration/issues/256)


squioc (approved):
LGTM

PR #1143, opened 2024-10-24
HarfangLab - add more actions

lvoloshyn-sekoia:


HarfangLab/harfanglab/base.py
```diff
@@ -38,3 +41,27 @@ def get_url(self, arguments):
...
             path += f"?{'&'.join(query_arguments)}"
         return path
+
+    def json_argument(self, name: str, arguments: dict, required: bool = True) -> Any:
+        """Get a JSON Argument by direct reference of by reading a file.
+
+        If `name` is inside arguments, returns the value.
+        If `name`_path is inside arguments, returns the content of the file
+        """
+        # @todo doesn't work with booleans by default - fix the error in Sekoia SDK
```
squioc, line 51: :disappointed:

HarfangLab/harfanglab/iocs.py
```diff
@@ -0,0 +1,173 @@
...
+                if "x_ic_observable_types" in object.keys():
+                    result["description"] = self.get_reference_url(object["id"])
+
+                # Compare expiration data if exists
+                valid_until = object.get("valid_until")
+                if valid_until:
+                    current_datetime = datetime.now(timezone.utc)
+                    valid_until_datetime = isoparse(valid_until)
+                    if valid_until_datetime >= current_datetime:
+                        result["expired"].append(result)
```
squioc, line 124: I think it's the `results` that is expected here, and not the current result:
```suggestion
                        results["expired"].append(result)
```

squioc (changes requested):
Thanks for the job.
One fix to apply to the PR.

HarfangLab/harfanglab/iocs.py
lvoloshyn-sekoia: You're right, thanks

HarfangLab/harfanglab/iocs.py
```diff
@@ -0,0 +1,173 @@
...
+                # Add a direct link in description if the data is originating from Sekoia.io
+                if "x_ic_observable_types" in object.keys():
+                    result["description"] = self.get_reference_url(object["id"])
+
+                # Compare expiration data if exists
+                valid_until = object.get("valid_until")
+                if valid_until:
+                    current_datetime = datetime.now(timezone.utc)
+                    valid_until_datetime = isoparse(valid_until)
+                    if valid_until_datetime >= current_datetime:
```
squioc, line 123: The comparison is inverted:
```suggestion
                    if valid_until_datetime < current_datetime:
```

HarfangLab/harfanglab/iocs.py
```diff
@@ -0,0 +1,173 @@
...
+                ids_to_remove.extend(found_ids)
+
+        if len(ids_to_remove) > 0:
+            self.delete_indicator_by_ids(indicator_ids=ids_to_remove)
+
+    def create_indicators(self, indicators: list) -> None:
+        if len(indicators) > 0:
+            self.log(f"Pushing {len(indicators)} new indicators to HarfangLab", level="info")
+            for ioc in indicators:
+                self.api_create_indicator(indicator=ioc)
```
squioc, line 154: It seems that the API doesn't support to create an IOC that already exist. We got the following error:
```
ERROR - status_code=400 reason="Bad Request" error="b'{\"__all__\":[\"Ioc rule with this Type, Value and Source already exists.\"]}'" event="Failed request to HarfangLab API" logger=harfanglab.iocs level=error timestamp=2024-10-28T16:03:12.006850Z
```
I think we should check if the indicator already exist

squioc (changes requested):
Some issues detected in the IOC creation.

HarfangLab/harfanglab/iocs.py
lvoloshyn-sekoia: 😅 you're right, thanks

HarfangLab/harfanglab/iocs.py
```diff
@@ -0,0 +1,193 @@
...
+
+        if len(ids_to_remove) > 0:
+            self.delete_indicator_by_ids(indicator_ids=ids_to_remove)
+
+    def create_indicators(self, indicators: list) -> None:
+        if len(indicators) > 0:
+            self.log(f"Pushing {len(indicators)} new indicators to HarfangLab", level="info")
+            for ioc in indicators:
+                try:
+                    self.api_create_indicator(indicator=ioc)
```
squioc, line 158: With `_handle_response_error`, if we face an 400 response, we will log an error then an info about the fact that the IoC exist. This will mislead the customer. I think we should move the call to `_handle_response_error` in the exception block below

squioc (changes requested):

squioc (approved):
LGTM. Thanks for the work

PR #1146, opened 2024-10-28
add search module in AD

pbivic:


MicrosoftActiveDirectory/action_search.json
```diff
@@ -0,0 +1,41 @@
...
+        "description": "The starting point an LDAP server uses when searching for users authentication within your Directory. (e.g DC=example-domain,DC=com)",
+        "type": "string"
+      },
+      "attributes": {
+        "title": "Attributes",
+        "description": "Attributes you want to retrieve (default will be ALL) ",
+        "type": "array",
+        "items": {
+          "type": "string"
+        }
```
squioc, line 26: In order to express that the property can be nullable:
```suggestion
        "anyOf": [
          {
            "type": "array",
            "items": {
              "type": "string"
            }
          },
          {"type": "null"}
        ],
        "default": null
```

squioc (changes requested):
LGTM. A little change to apply in the definition of the action.

MicrosoftActiveDirectory/tests/test_search.py
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    return a
+
+
+@pytest.fixture
+def one_user_dn():
+    return [["CN=integration_test,CN=Users,DC=lab,DC=test,DC=com", 512]]
+
+def test_search():
```
TOUFIKIzakarya, line 24: I'm not sure that this test will pass. I think you need to mock the search method.

TOUFIKIzakarya (commented):
Good job paul. Thanks for that.
just a little remark about tests

TOUFIKIzakarya (approved):
LGFM

squioc (approved):
LGTM

PR #1148, opened 2024-10-29
VadeCloud: Stop connector on authentication errors

squioc:
Log as critical, any authentication error. This will stop the connector after 5 failures


TOUFIKIzakarya (approved):
LGFM

PR #1151, opened 2024-10-30
Tehtris: change the fitler for duplicated events

squioc:
The `uid` is not unique and discard too many events. Use `id` instead.


TOUFIKIzakarya (approved):
LGFM

PR #1156, opened 2024-11-04
Office365: Reduce to 7 days the catch back in the past

squioc:
Reduce to 7 days the catch back in the past as the API only support 7 days.
```
{"error":{"code":"AF20055","message":"Start time and end time must both be specified (or both omitted) and must be less than or equal to 24 hours apart, with the start time prior to end time and start time no more than 7 days in the past. StartTime:2024-10-05T10:27:13.619091+00:00, EndTime:2024-10-05T10:57:13.619091+00:00"}}
```


TOUFIKIzakarya (approved):
LGFM

PR #1182, opened 2024-11-19
Add Trend Micro Vision One connector

lvoloshyn-sekoia:


squioc (commented):
Thanks for the work.
Can you update the version number in the manifest file of the module and add an entry in the CHANGELOG.md?

PR #1184, opened 2024-11-20
Sophos: add ratelimit per day

squioc:
- Add the ratelimit per day
- Increase the frequency to 5 minutes for the XDR connector


TOUFIKIzakarya (approved):
LGFM

PR #1185, opened 2024-11-22
PaloAltoCortexXDR: change the way to handle responses

squioc:
- Fix the way to handle the responses
- Handle HTTP errors
- Stop the connector when credentials or permissions are invalids


TOUFIKIzakarya (approved):
LGFM, thanks for the fix

PR #1190, opened 2024-11-27
Retrieve only attributes for search AD command

penhouetp:


TOUFIKIzakarya:
Hello Pierre, 
Thanks for the fix, can you please add an entry to changelog and update the version in the manifest file.

TOUFIKIzakarya (approved):
LGFM

squioc (approved):
LGTM

PR #1194, opened 2024-11-28
chore: rename asset CRUD actions to differentiate v1 from v2

ysindel:


squioc (changes requested):
Can you increase the version number in the manifest.json? Thank you

squioc (approved):
Don't care about the previous comment.

PR #1201, opened 2024-12-03
Trend Micro Workbench Alerts - fix naming

lvoloshyn-sekoia:


squioc (changes requested):
Can you add an entry in the CHANGELOG.md? Thank you

squioc (approved):

PR #1205, opened 2024-12-03
fix create asset url

ysindel:


Darkheir (commented):
Won't this PR break existing playbooks using those actions ?

Darkheir (approved):

PR #1209, opened 2024-12-04
SentinelOne: fix module configuration

squioc:
Set the hostname and the api token as required in the module configuration.


TOUFIKIzakarya (commented):
LGFM but there's a problem in the `test_consume.py` tests

TOUFIKIzakarya (approved):
LGFM

PR #1211, opened 2024-12-09
Get cases with get alert action

aheuze:


Sekoia.io/action_get_an_alert.json
```diff
@@ -15,6 +15,12 @@
       "stix": {
         "in": "query",
         "type": "boolean"
+      },
+      "cases": {
+        "in": "query",
+        "type": "boolean",
+        "description": "Get the cases associated with this alert",
+        "default": true
```
squioc, line 23: I think that the philosophy of this boolean is to not return the cases associated to an alert by default.
```suggestion
        "default": false
```

squioc (commented):
LGTM.
Just a little discussion about the boolean

squioc (approved):
LGTM

PR #1213, opened 2024-12-09
Add action AddEventsToACase

aheuze:


Sekoia.io/sekoiaio/operation_center/__init__.py
```diff
@@ -297,6 +297,15 @@
     },
 )
 
+AddEventsToACase = type(
+    "AddEventsToACase",
+    (GenericAPIAction,),
+    {
+        "verb": "post",
+        "endpoint": base_url + "cases/{case_uuid}/events",
```
squioc, line 305: The variable must have the same name as in the action declaration.
```suggestion
        "endpoint": base_url + "cases/{uuid}/events",
```

squioc (changes requested):
Thanks for the contribution.
A little fix to apply

PR #1227, opened 2024-12-12
Fix: Azure avoid duplicates (316)

vg-svitla:
Possible fix for [316](https://github.com/SekoiaLab/integration/issues/316)


squioc:
I close this PR as the issue was not related to the connector, and the proposed fix will change nothing.

PR #1229, opened 2024-12-13
SentinelOne: fix the pagination

squioc:
- Reduce the verbosity of the GQL client in the logs
- Fix the pagination in the Singularity connector.


TOUFIKIzakarya (approved):
LGFM

PR #1234, opened 2024-12-18
Add several actions on Cases

aheuze:
Add the following actions:
- Create case
- Modify case
- Get case
- Add a comment to a case


squioc (approved):
LGTM

PR #1242, opened 2025-01-07
Chore: Upgrade projects

squioc:
Upgrade projects in order that they are compatible with Poetry 2.0.


vg-svitla (approved):

r1chev:
@squioc it seems good but FYI you've also added some extra files (python, tests) not related with the poetry updates

squioc:
@r1chev good point. I fixed it.

PR #1243, opened 2025-01-07
feat(Triage) option to exclude binary with suspicious analysis

r1chev:
Adding an other option `exclude_suspicious_analysis` (except for linux), to filter out some reports based on either:                                 
* the number of behavioral dynamic analysis (less than the default number).
* a score gap (with a standard deviation > 2.5) between two behavioral analysis.

We want to detect reports where user is using the sandbox to run some suspicious behavior not related to the first sample submitted (_.i.e._ web browsing session)


r1chev:
Thx @squioc for your help!

Triage/trigger_triage.json
```diff
@@ -20,6 +20,10 @@
       "exclude_signed": {
         "type": "boolean",
         "description": "Do not retrieve report with signed binary"
+      },
+      "exclude_suspicious_analysis": {
+        "type": "boolean",
+        "description": "Do not retrieve report with only one behavioral or with a score gap between two dynamic analysis"
```
Darkheir, line 26:
```suggestion
        "description": "Do not retrieve report with only one behavioral or with a score gap between two dynamic analysis",
        "default": false
```

Darkheir (commented):
Code looks good to me, but the version of the module should be bumped in the `manifest.json` file.

r1chev:
> Code looks good to me, but the version of the module should be bumped in the `manifest.json` file.

Thx done!

Darkheir (approved):

PR #1254, opened 2025-01-16
support list in alert trigger

rombernier:
Add support for list in rule filter for alert trigger, remove support for rule uuid, as it's must not be used to avoid conflict between rule uuid and rule instance uuid

### DIFF
--- a/Sekoia.io/CHANGELOG.md
+++ b/Sekoia.io/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2025-02-10 - 2.67.4
+
+### Added
+
+- Update Alert trigger with rule_names_filter
+ 
 ## 2025-01-27 - 2.67.3
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


squioc (commented):
I'm not sure about my proposition, and I think we have to check with a frontend developer how the interface will react with this change.

Carole-Besset:
> Hello @Carole-Besset, do you know if this change will work with the front ?

Yes it'll have some impacts on front-end on the playbook configuration part, as we're reading the "type" value as a string, not an array (and we're doing a switch on it to display the right component : object, boolean array..)

rombernier:
Hello @gaelmuller, I update the PR to follow the option with a new filter, what do you think ?

gaelmuller (approved):
Looks good to me, but I think you should specify the type of items inside the array

rombernier:
Hello @squioc, can we specified array item type in trigger json schema, I didn't find any example ? :)

squioc:
Hello @rombernier, I will be able to find an example [here](https://github.com/SEKOIA-IO/automation-library/blob/main/Cybereason/trigger_pull_events_new.json#L27-L33).

rombernier:
Hello @squioc, fix is done, can you approve and merge ? Thanks

squioc (approved):
LGTM

PR #1261, opened 2025-01-24
Fix: Mimecast verify that fetch date is after then 7 days ago

vg-svitla:


Mimecast/mimecast_modules/connector_mimecast_siem.py
```diff
@@ -69,7 +69,7 @@ def most_recent_date_seen(self):
         most_recent_date_seen = isoparse(most_recent_date_seen_str)
 
         # We don't retrieve messages older than one day
-        one_day_ago = now - timedelta(days=7)
+        one_day_ago = now - timedelta(days=1)
```
squioc, line 72: I think that the name of the variable (and the comment upper are wrong) and that we must keep 7 days here.

squioc (changes requested):

squioc (approved):
LGTM

PR #1268, opened 2025-01-30
Hotfix: Checkpoint Harmony token expiration

vg-svitla:
Closes [389](https://github.com/SekoiaLab/integration/issues/389)


Checkpoint/client/token_refresher.py
```diff
@@ -109,3 +115,39 @@ async def get_token(self) -> RefreshedToken[CheckpointToken]:
...
+        Get access token.
+
+        Yields:
+            RefreshedTokenT:
+        """
+        cached_token: RefreshedToken[CheckpointToken] | None = await _cache.get(self.cache_key)
+        if cached_token is None:
+            cached_token = await self.get_token()
+            ttl = cached_token.ttl - 1  # 1 second less to be sure that token is expired
+            if ttl >= 1:
```
squioc, line 131: Do you know how long last the tokens?
Can we increase the safety period to 30s?

squioc (commented):
Thank you for the job.

Checkpoint/client/token_refresher.py
vg-svitla: To be honest TTL is inside of the response. However I think we can set it to 30 seconds. Why not

vg-svitla: I'll double check and provide a fix

TOUFIKIzakarya (approved):
LGFM

squioc (approved):
LGTM

PR #1269, opened 2025-01-31
Mimecast - fix cursor

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/392


Mimecast/mimecast_modules/connector_mimecast_siem.py
squioc, line 1: For a smooth transition, we must read the date of the last collected event to restart the collect at the correct day after having deployed the connector in production.

squioc (changes requested):
Thanks for the fix.
However, for a smooth transition, we need to read the date of the last collected event.

Mimecast/tests/test_mimecast_siem_logs.py
```diff
@@ -263,3 +245,51 @@ def test_permission_denied(
...
+        end_time = start_time + batch_duration
+        mock_time.time.side_effect = [start_time, end_time, end_time]
+
+        consumer = MimecastSIEMWorker(connector=trigger, log_type="process", client=api_client)
+        assert consumer.old_cursor == fake_date
+
+        consumer.next_batch()
+
+        assert consumer.old_cursor is None
+        assert consumer.cursor.offset == "tokenNextPageLast=="
```
squioc, line 295: Can you check that the first request, to the Mimecast endpoint, used the date from the cursor?  You can use the [history](equests-mock.readthedocs.io/en/latest/history.html) from the mock.
Thank you

Mimecast/mimecast_modules/connector_mimecast_siem.py
```diff
@@ -52,73 +57,79 @@ def stop(self):
...
+            logger.info("Starting with old datetime cursor", datetime=self.old_cursor.isoformat())
+            params["dateRangeStartsAt"] = self.old_cursor.strftime("%Y-%m-%d")
+
+        elif self.cursor.offset is None:
+            # provide date range to start
+            start_at = datetime.today()
+            params["dateRangeStartsAt"] = start_at.strftime("%Y-%m-%d")
+
+        else:
+            params["nextPage"] = self.cursor.offset
```
squioc, line 109: After the first restart, is `old_cursor` erased from the context?
However, I propose to reverse the order of the tests:
- if `cursor.offset` is None and `old_cursor` is not None, then we use `old_cursor`
- elif `cursor.offset` is not None, then we use `cursor.offset`
- else we use the today date as the starting date.

Alternatively, we delete the `old_cursor` after the first request (maybe risky).

squioc (changes requested):

squioc (approved):
LGTM. Thank you for the fix

PR #1273, opened 2025-02-04
add support for file input

rombernier:


Sekoia.io/sekoiaio/operation_center/synchronize_assets_with_ad.py
```diff
@@ -1,24 +1,33 @@
...
     """
 
-    def run(self, arguments: Arguments):
-        user_ad_data = arguments.user_ad_data
+    def run(self, arguments: Arguments) -> List[Dict[str, Any]]:
+        # Determine if user_ad_data is a list
+        if arguments.user_ad_file is not None:
+            user_ad_data = self.json_argument("user_ad_file", arguments)
+        else:
+            user_ad_data = arguments.user_ad_data
```
squioc, line 26: `json_argument` works with dictionary, not object. So, it won't work here.
Sadly, the simplest thing to do here is to convert back `arguments` to a dict :(

```suggestion
    def run(self, arguments: dict) -> List[Dict[str, Any]]:
        user_ad_data = self.json_argument("user_ad_data", arguments)
```

Sekoia.io/sekoiaio/operation_center/synchronize_assets_with_ad.py
```diff
@@ -143,33 +157,35 @@ def merge_assets(destination: str, sources: List[str]) -> None:
...
-        }
-        return response
+            response = {
+                "found_assets": list(found_assets),
+                "created_asset": created_asset,
+                "destination_asset": destination_asset,
+            }
+            responses.append(response)
+
+        return responses
```
squioc, line 191: It changes the output signature of the action. That may lead to break existing playbooks.
Maybe a better option is to duplicate this action and change the argument and the output according to the current change in this PR.

squioc (changes requested):
Thank you for the work.
However, I detected several issues in the PR.

Sekoia.io/sekoiaio/operation_center/synchronize_assets_with_ad.py
rombernier: Existing playbooks with this actions are all under our control.

Sekoia.io/sekoiaio/operation_center/synchronize_assets_with_ad.py
rombernier: Done

rombernier:
Hello @squioc, changes are done, can you check and approve ?

squioc (approved):
LGTM

Sekoia.io/pyproject.toml
```diff
@@ -15,6 +15,7 @@ ujson = "*"
 websocket-client = "*"
 gevent = "*"
 orjson = "*"
+mypy = "^1.15.0"
```
squioc, line 18: Can you remove this dependency? Thank you
```suggestion
```

squioc (changes requested):

squioc (approved):
LGTM

PR #1275, opened 2025-02-05
Add docs for Eset actions

lvoloshyn-sekoia:


squioc (changes requested):
Thank you for the document.
However, it seems that images are missing from the PR.

squioc (approved):
LGTM. Thank you

PR #1286, opened 2025-02-12
Improve BeyondTrust connector

lvoloshyn-sekoia:
- Change default batch frequency
- Use cache for session ids


squioc (approved):
LGTM. Thank you for the fix.

PR #1305, opened 2025-03-04
Fix: AWS Connector Memory Leak Issue (465)

vg-svitla:
Closes [465](https://github.com/SekoiaLab/integration/issues/465)

### DIFF
--- a/AWS/connectors/s3/__init__.py
+++ b/AWS/connectors/s3/__init__.py
@@ -176,6 +176,7 @@ class AbstractAwsS3QueuedConnector(AbstractAwsConnector, metaclass=ABCMeta):
             if not records:
                 continue_receiving = False
 
-        result += len(await self.push_data_to_intakes(events=records))
+        if records:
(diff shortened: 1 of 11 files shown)


AWS/connectors/s3/__init__.py
```diff
@@ -162,15 +163,19 @@ async def next_batch(self, previous_processing_end: float | None = None) -> tupl
...
                             message=f"Failed to fetch content of {record}: {str(e)}",
                             level="warning",
                         )
 
-            if len(records) >= self.limit_of_events_to_push or not records:
+            if not records:
                 continue_receiving = False
 
-        result = await self.push_data_to_intakes(events=records)
+        result += len(await self.push_data_to_intakes(events=records))
```
squioc, line 179: Maybe we can add a condition to check if `records` is empty or not and avoid calling `push_data_to_intakes` for empty list.

squioc:
Thnak you for the fix.
I made a comment on the fix.

PR #1313, opened 2025-03-06
Fix: Wiz Stop Connector only on 401 and 403

vg-svitla:


squioc (changes requested):

squioc (approved):
LGTM. Thank you for the improvement.

PR #1314, opened 2025-03-10
Fix cache problem

TOUFIKIzakarya:
Main issue : [Issue](https://github.com/SekoiaLab/integration/issues/455)

## Summary by Sourcery

Fixes an issue where events were not being properly cached, leading to potential duplication of events. It also updates the connector to use a cursor for tracking the last event date, improving the reliability of event collection.

Bug Fixes:
- Fixes a caching problem that could lead to duplicate events being ingested.

Enhancements:
- Updates the connector to use a cursor for tracking the last event date, improving the reliability of event collection.

Tests:
- Adds a persistent storage fixture to the tests.

### DIFF
--- a/SentinelOne/sentinelone_module/helpers.py
+++ b/SentinelOne/sentinelone_module/helpers.py
@@ -85,7 +85,7 @@ def stix_to_indicators(stix_object, supported_types_map):
     return results
 
 
-def filter_collected_events(events: Sequence, getter: Callable, cache: Cache, context: PersistentJSON ) -> list:
+def filter_collected_events(events: Sequence, getter: Callable, cache: Cache, context: PersistentJSON) -> list:
     """
     Filter events that have already been filter_collected_events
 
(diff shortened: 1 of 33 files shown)


squioc (changes requested):
Can you load and save the cache only when needed? i.e, loading the cache at the starting and save it after each btach of events fetched.

SentinelOne/sentinelone_module/helpers.py
```diff
@@ -8,6 +8,8 @@
 from cachetools import Cache
 from stix2patterns.pattern import Pattern
 
+from sekoia_automation.storage import PersistentJSON
+
```
squioc, line 12: It seems that this import is useless. I suggest removing it:
```suggestion
```

squioc (changes requested):
Thank you for this fix.
A little change to apply before we can merge the PR.

squioc (approved):
LGTM. Thank you for the work.

PR #1316, opened 2025-03-12
Mimecast: change the way to consume S3 objects

squioc:
Use generators, instead of lists, to consume the content of the S3 objects returned by the API.
The events are consumed in batch of 10000 items by default.
This behavior should fix the memory issue observed on the connector.

## Summary by Sourcery

Bug Fixes:
- Fix memory issues by using generators to process S3 objects, which prevents loading large lists of events into memory at once.


squioc:
I close this PR in favor of https://github.com/SEKOIA-IO/automation-library/pull/1318

PR #1318, opened 2025-03-13
Mimecast: change the way to consume S3 objects

squioc:
Use generators, instead of lists, to consume the content of the S3 objects returned by the API.
The events are consumed in batches of 10000 items by default.
This behavior should fix the memory issue observed on the connector.

## Summary by Sourcery

Refactor the connector to use generators instead of lists to consume S3 objects, addressing a memory issue. The events are now consumed in batches of 10000 items by default.

Bug Fixes:
- Fix memory issues by using generators instead of lists when fetching events.
- Fix the pagination

Enhancements:
- Use generators instead of lists to consume the content of the S3 objects returned by the API.
- Consume events in batches of 10000 items by default.


TOUFIKIzakarya (approved):
Thanks for the fix.
LGFM

PR #1319, opened 2025-03-14
Fix: aws oom issue

vg-svitla:
## Summary by Sourcery

Improves memory management by processing events in batches and using async methods to avoid unnecessary memory allocation when parsing events from S3 buckets.


squioc (changes requested):
Thank you for the proposal. However, IMHO, the fix misses the point about the decompression, in memory, of Gzipped S3 Objects.

A counter-proposal was made: https://github.com/SEKOIA-IO/automation-library/pull/1328

squioc:
I close this PR as the alternative one was merged.
Thanks again for the proposal.

PR #1325, opened 2025-03-20
Zscaler - add logging

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/504

Seems to be authorization error due to wrong credentials. But we have to fix logging to make it clear

## Summary by Sourcery

Improve logging for authentication failures and invalid indicators.

### DIFF
--- a/Zscaler/CHANGELOG.md
+++ b/Zscaler/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2025-03-27 - 1.0.1
+
+### Fixed
+
+- Fixed logging for authorization errors 
+
 ## 2024-05-28 - 1.0.0
 
 ### Changed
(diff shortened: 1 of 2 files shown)


Zscaler/zscaler/block_ioc.py
```diff
@@ -10,24 +10,27 @@
...
+                cloud_name=self.module.configuration["base_url"],
                 api_key=self.module.configuration["api_key"],
                 username=self.module.configuration["username"],
                 password=self.module.configuration["password"],
             )
             return api
+
         except Exception as e:
-            print(f"ZIA authentication failed: {str(e)}")
+            self.log(f"ZIA authentication failed", level="critical")
```
squioc, line 22:
```suggestion
            self.log("ZIA authentication failed", level="critical")
```

Zscaler/zscaler/block_ioc.py
```diff
@@ -10,24 +10,27 @@
...
+            self.log_exception(e)
             return None
```
squioc, line 24: Instead of return None (that generated the error in the issue), I suggest raising the caught exception.

squioc (changes requested):
Thank you for the work.
Some suggestions were proposed

squioc (changes requested):
Can you bump the version number in the manifest and add an entry in the CHANGELOG.md? Thank you

squioc (approved):
LGTM. Thank you for the fix.

PR #1336, opened 2025-04-02
update output type

rombernier:
## Summary by Sourcery

Update the output type for the synchronize asset action to return a consistent data structure

Bug Fixes:
- Fixed the output type for the synchronize asset action to return a dictionary with a 'data' key containing the responses

Chores:
- Bumped version from 2.67.9 to 2.67.10 in manifest.json
- Updated CHANGELOG.md with details of the fix


squioc (approved):
LGTM.

PR #1337, opened 2025-04-02
Fix: TrendMicro fix (486)

vg-svitla:
Possible fix for [486](https://github.com/SekoiaLab/integration/issues/486)

## Summary by Sourcery

Improve event handling and caching mechanism for TrendMicro Email Security connector to prevent duplicate event processing

Bug Fixes:
- Prevent duplicate event processing by tracking recently pushed events in local cache

Enhancements:
- Add methods to get and set recent pushed events
- Modify event fetching logic to skip already processed events

Chores:
- Update version to 1.6.1
- Add pragmas for code coverage exclusion

### DIFF
--- a/TrendMicro/trendmicro_modules/trigger_email_sec.py
+++ b/TrendMicro/trendmicro_modules/trigger_email_sec.py
@@ -3,7 +3,7 @@ import traceback
 import urllib.parse
 from functools import cached_property
 from threading import Event, Lock, Thread
-from typing import Generator, Any
+from typing import Any, Generator
 
 import orjson
 from pydantic import Field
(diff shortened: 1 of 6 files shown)


TrendMicro/trendmicro_modules/trigger_email_sec.py
```diff
@@ -76,11 +76,33 @@ def get_last_timestamp(self) -> int:
...
+        self.connector.context_lock.release()
+
+    def set_recent_pushed_events(self, events: list[Any]) -> None:
+        self.connector.context_lock.acquire()
+
+        with self.connector.context as cache:
+            if self.log_type not in cache.keys():
+                cache[self.log_type] = {}
+
+            cache[self.log_type]["recent_pushed_events"] = events
```
squioc, line 105: You save the list of events with all information. In order to reduce the pressure on the cache, can you only save the `action` and `mailID` from the events, then use this information in the filter?

squioc (commented):
Thank you for the fix.
In order to reduce the pressure of the cache, can you focus the filtering on the `mailID` and the `action` of the events?

TrendMicro/trendmicro_modules/trigger_email_sec.py
vg-svitla: I also propose to add `messageID`

squioc (approved):
LGTM

PR #1340, opened 2025-04-03
Adding render filter for get case

LenaigKaliou:


squioc (changes requested):
Could you bump the version number in the manifest?

squioc (approved):
LGTM

PR #1341, opened 2025-04-07
Fortigate: change the output of the action that add new Address group

squioc:
- Lists are not supported as output of the `run` method. Wrap the returned list into a dict.

## Summary by Sourcery

Bug Fixes:
- Fix the output of the 'Add new Address group to a Fortigate firewall' action by wrapping the returned list into a dictionary


TOUFIKIzakarya (approved):
LGFM

PR #1345, opened 2025-04-07
Fix on cases triggers

LenaigKaliou:
## Summary by Sourcery

Refactor and improve the case triggers filtering mechanism in the Sekoia.io integration

Bug Fixes:
- Fix filtering logic for different case trigger events to ensure consistent filtering across different event types

Enhancements:
- Split the monolithic _filter_notifications method into separate filtering methods for mode, priority, assignees, and case UUIDs
- Improve the filtering logic by separating concerns and making the code more modular

Tests:
- Update test cases to use a new case alerts updated trigger fixture
- Modify test cases to work with the new filtering approach

### DIFF
--- a/Sekoia.io/sekoiaio/triggers/cases.py
+++ b/Sekoia.io/sekoiaio/triggers/cases.py
@@ -18,39 +18,34 @@ class SecurityCasesTrigger(_SEKOIANotificationBaseTrigger):
         ("case", "alerts-updated"),
     ]
 
-    def _filter_by_mode(self, message) -> bool:
-        case_attrs = message.get("attributes", {})
+    def _filter_by_mode(self, case) -> bool:
         mode_filter = self.configuration.get("mode_filter")
-        if mode_filter and case_attrs.get("manual") != (mode_filter == "manual"):
+        if mode_filter and case.get("manual") != (mode_filter == "manual"):
             return False
 
         return True
 
-    def _filter_by_priority(self, message) -> bool:
-        case_attrs = message.get("attributes", {})
+    def _filter_by_priority(self, case) -> bool:
         priority_uuids_filter = self.configuration.get("priority_uuids_filter")
-        if priority_uuids_filter and case_attrs.get("custom_priority_uuid") not in priority_uuids_filter:
+        if priority_uuids_filter and case.get("custom_priority_uuid") not in priority_uuids_filter:
             return False
 
         return True
 
-    def _filter_by_assignees(self, message) -> bool:
-        case_attrs = message.get("attributes", {})
+    def _filter_by_assignees(self, case) -> bool:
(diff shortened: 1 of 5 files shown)


Sekoia.io/sekoiaio/triggers/cases.py
```diff
@@ -150,6 +161,12 @@ def handle_event(self, message):
         case_attrs = message.get("attributes", {})
         event_type: str = message.get("type", "")
         event_action: str = message.get("action", "")
+        filter_notifications = [
+            self._filter_by_mode(message),
+            self._filter_by_priority(message),
+            self._filter_by_assignees(message),
+            self._filter_by_uuids(message),
+        ]
```
squioc, line 169: Filters must be applied on the case, not on the notification.
Can you move this after you retrieve the case from the API and update the `filter_by_*` methods to work with the [response](https://docs.sekoia.io/xdr/develop/rest_api/alert/#tag/Cases/operation/get_case_resource) of the API?

Sekoia.io/sekoiaio/triggers/cases.py
```diff
@@ -205,6 +222,12 @@ def handle_event(self, message):
...
```
squioc, line 230: Same here

Sekoia.io/sekoiaio/triggers/cases.py
```diff
@@ -101,6 +106,11 @@ def handle_event(self, message):
         case_attrs = message.get("attributes", {})
         event_type: str = message.get("type", "")
         event_action: str = message.get("action", "")
+        filter_notifications = [
+            self._filter_by_mode(message),
+            self._filter_by_priority(message),
+            self._filter_by_assignees(message),
+        ]
```
squioc, line 113: Same here.

Sekoia.io/sekoiaio/triggers/cases.py
```diff
@@ -18,36 +18,41 @@ class SecurityCasesTrigger(_SEKOIANotificationBaseTrigger):
...
             return False
 
-        # Filter by assignees
+        return True
+
+    def _filter_by_assignees(self, message) -> bool:
+        case_attrs = message.get("attributes", {})
         assignees_filter = self.configuration.get("assignees_filter")
         if assignees_filter:
             case_details = self._retrieve_case_from_caseapi(case_attrs.get("uuid"))
```
squioc, line 41: The filter must be applied on the case from the API. So, this line is useless, as you must already have fetched the case.

squioc (changes requested):
Thank you for splitting the filters. However, the filters need to be applied to the case retrieved from the API, not the notification.

squioc (approved):
LGTM. Thank you for the changes

PR #1346, opened 2025-04-10
Sekoia.io: fix cases

squioc:
- Add more tests
- Fix the name of the delete case action
- Fix the name of the list of deleted alerts in the case alerts updated trigger.


LenaigKaliou (approved):
LGTM

PR #1350, opened 2025-04-18
Sekoia GetAssets v2 - fix

lvoloshyn-sekoia:
[According to the docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat), `datetime.fromisoformat` won't parse datetime like `2011-11-04T00:05:23Z` (specifically with "Z" in the end) in python version < 3.11

We allow python 3.10 in Sekoia module, thus getting an error. I believe the action is failing due to this. I replaced the parsing function with a stable analog from `dateutil`.

https://github.com/SekoiaLab/integration/issues/564

## Summary by Sourcery

Fix datetime parsing compatibility for Python versions before 3.11

Bug Fixes:
- Replace `datetime.fromisoformat()` with `dateutil.parser.isoparse()` to correctly handle ISO datetime strings with 'Z' timezone in Python 3.10

Enhancements:
- Improve datetime parsing robustness across different Python versions

Build:
- Add python-dateutil dependency to project dependencies


TOUFIKIzakarya (approved):
LGFM

PR #1361, opened 2025-05-14
Sekoia GetAssetsV2 - remove datetime parsing at all

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/564

If we use datetime only to dump it later, we don't need to parse it at all.

## Summary by Sourcery

Stop parsing datetime fields in the GetAsset action and treat created_at and updated_at as raw strings

Enhancements:
- Remove datetime parsing logic and related imports from the GetAssetV2 implementation

Build:
- Drop the python-dateutil dependency from the project

Tests:
- Update asset tests to expect created_at and updated_at as ISO-format strings instead of datetime objects

### DIFF
--- a/Sekoia.io/poetry.lock
+++ b/Sekoia.io/poetry.lock
@@ -211,13 +211,13 @@ virtualenv = ["virtualenv (>=20.0.35)"]
 
 [[package]]
 name = "cachecontrol"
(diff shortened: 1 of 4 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #1362, opened 2025-05-14
Fix: sentinel one custom errors (558)

vg-svitla:
Closes [558](https://github.com/SekoiaLab/integration/issues/558)

## Summary by Sourcery

Improve error handling in SentinelOne logs connector by introducing custom exceptions for empty and error-filled API responses, updating consumers to validate and raise these errors, refining logging in next_batch, and updating documentation, tests, and versioning to 1.20.6

Enhancements:
- Introduce SentinelOneManagementResponseError and SENTINEL_ONE_EMPTY_RESPONSE to enforce error handling for empty or erroneous API responses
- Update pull_events in activity and threat consumers to validate responses and raise custom exceptions before processing
- Add specific exception handlers in next_batch to log contextual errors from SentinelOne API
- Rename event identifier variable from events_id to events_ids for consistency

Documentation:
- Reformat argument table in CONFIGURE.md and update version to 1.20.6 in manifest and CHANGELOG.md

Tests:
- Add test to assert SentinelOneManagementResponseError is raised when API responses contain errors

Chores:
- Bump package version to 1.20.6 and add release notes to CHANGELOG

### DIFF
--- a/SentinelOne/sentinelone_module/logs/connector.py
+++ b/SentinelOne/sentinelone_module/logs/connector.py
@@ -17,11 +17,7 @@ from sekoia_automation.checkpoint import CheckpointDatetime
 from sekoia_automation.connector import Connector
 
 from sentinelone_module.base import SentinelOneModule
-from sentinelone_module.exceptions import (
-    SENTINEL_ONE_EMPTY_RESPONSE,
-    SentinelOneError,
-    SentinelOneManagementResponseError,
-)
+from sentinelone_module.exceptions import SENTINEL_ONE_EMPTY_RESPONSE, SentinelOneManagementResponseError
 from sentinelone_module.helpers import filter_collected_events
 from sentinelone_module.logging import get_logger
 from sentinelone_module.logs.configuration import SentinelOneLogsConnectorConfiguration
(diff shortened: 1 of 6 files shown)


SentinelOne/sentinelone_module/logs/connector.py
```diff
@@ -218,17 +237,26 @@ def __init__(self, connector: "SentinelOneLogsConnector"):
...
+        events_ids = []
         while self.running:
             # Fetch threats
-            threats = self.management_client.client.get(endpoint="threats", params=query_filter.filters)
+            threats: ManagementResponse | None = self.management_client.client.get(
+                endpoint="threats", params=query_filter.filters
+            )
+
+            if threats is None:
+                raise SENTINEL_ONE_EMPTY_RESPONSE
```
squioc, line 256: I propose breaking the loop instead of raising an error:
- a nullable or an empty `threats` can be something expected from the API
- a new exception is raised, we lost the content of `events_ids`

squioc (changes requested):
Thank you for the fix. I proposed some changes.

SentinelOne/sentinelone_module/logs/connector.py
vg-svitla: I did some renaming so it is more clear now.
Basically `threats` it is not the data at this point of time. It is exactly the response with next type `ManagementResponse | None`
But I agree with suggestion for check nullability of `data`, so I pushed the update

squioc (approved):
LGTM

PR #1370, opened 2025-05-23
Fix return type

penhouetp:
## Summary by Sourcery

Change get_agent_telemetry to return a dict containing the results under a "data" field, update tests and documentation, and bump the module version to 1.26.1.

Enhancements:
- Wrap telemetry results in a dictionary under the "data" key instead of returning a raw list.
- Update the return type annotation of the get_agent_telemetry action to reflect the new structure.

Documentation:
- Add a changelog entry for version 1.26.1 describing the return type update.

Tests:
- Adjust integration tests to assert against response["data"] and its length.

Chores:
- Bump manifest version to 1.26.1.

### DIFF
--- a/HarfangLab/CHANGELOG.md
+++ b/HarfangLab/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #1386, opened 2025-06-05
fix(sekoiaio): returning empty list when fetching source references fa…

alexandreLavenant:
fix(sekoiaio): returning empty list when fetching source references failed

## Summary by Sourcery

Return an empty list when source reference fetches fail and release version 2.68.2

Bug Fixes:
- Return an empty list on failed source reference fetch instead of raising an error

Build:
- Bump Sekoia.io integration version to 2.68.2

Documentation:
- Add changelog entry for version 2.68.2


TOUFIKIzakarya (approved):
LGFM

PR #1390, opened 2025-06-11
Add slug to action

squioc:
Add slug to actions


StephKll3c (approved):
Such efficiency! Thank :)

PR #1395, opened 2025-06-12
Fix size of batch problem

TOUFIKIzakarya:
## Summary by Sourcery

Implement batching in the BlockMaliciousFilesAction to limit requests to 100 hashes per payload and return the total number of blocked files, update related tests and bump package version

Bug Fixes:
- Fix oversized hash list by splitting payloads into batches of up to 100 hashes

Enhancements:
- Refactor request_payload to yield batches of payloads and update run method to aggregate blocked file counts

Tests:
- Add fixture and test to verify handling of multiple STIX objects and batch count logic
- Remove unused imports and adjust tests to assert blocked_files_count

### DIFF
--- a/PaloAltoCortexXDR/cortex_module/actions/action_block_malicious_files.py
+++ b/PaloAltoCortexXDR/cortex_module/actions/action_block_malicious_files.py
@@ -110,7 +110,7 @@ class BlockMaliciousFilesAction(PaloAltoCortexXDRAction):
         if stix_objects is None or len(stix_objects) == 0:
             self.log("Received stix_objects were empty")
 
-        hashes : list[str] = []
+        hashes: list[str] = []
         for value in stix_objects:
             # Extract value and type from pattern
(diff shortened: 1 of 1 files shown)


PaloAltoCortexXDR/cortex_module/actions/action_block_malicious_files.py
```diff
@@ -122,16 +121,37 @@ def request_payload(self, arguments: dict[str, Any]) -> dict[str, Any]:
...
                     continue
 
+                if len(hashes) >= 100:
+                    yield {
+                        "request_data": {
+                            "hash_list": hashes,
+                            "comment": model.comment,
+                            "incident_id": model.incident_id,
+                        }
+                    }
```
squioc, line 131: I suggest clearing `hashes` after this yield, in order to not duplicate the indicators

TOUFIKIzakarya: Good catch, thanks for that.

PR #1398, opened 2025-06-16
Add Vectra module

lvoloshyn-sekoia:
## Summary by Sourcery

Introduce the Vectra module with a new connector to fetch and forward entity scoring events

New Features:
- Add connector to periodically fetch account and host scoring events from Vectra Respond UX and forward them to intakes
- Implement Vectra API client with OAuth2 authentication, rate limiting, and retry support
- Introduce TimeStepper utility to compute and iterate polling time windows with lag management

Enhancements:
- Add persistent and in-memory caching to filter out already processed events
- Expose Prometheus metrics for forwarded events count, forwarding duration, and event lag

Build:
- Add pyproject.toml with dependencies, formatting, and test configurations

Tests:
- Add tests for TimeStepper behavior and connector fetch_events functionality

Chores:
- Update manifest schema to require base_url, client_id, and client_secret and bump module version to 1.1.0


Vectra/vectra_modules/connector_vectra_entity_scoring.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def __get_request(self, *args, **kwargs) -> Response:
+        try:
+            response = self.client.get(*args, **kwargs)
+            response.raise_for_status()
+
+            return response
+
+        except requests.exceptions.HTTPError as err:
+            self.log_exception(err)
```
squioc, line 143: I could be interesting to extract the payload of the response and log it as we (and the customer) can get information about the reason of the error.

squioc (changes requested):
LGTM. Thank you for the work.

I think, for API errors, we can extract the body of the response to get more details about the reasons.

PR #1401, opened 2025-06-18
chore: rename glimps new actions now in GA

jeromefellus-sekoia:


squioc (changes requested):
Thanks for the PR.
Can you increase the version number in the manifest file and add an entry in the CHANGELOG?

jeromefellus-sekoia:
done thanks

PR #1405, opened 2025-06-26
fix(http): Fix schema in URL request to accept JSON Array for JSON field

goudyj:
With the previous schema we were not able to make a post with a JSON array like: `[{"foo": "bar"},{"foo": "bar"}]`

This PR fixes that.

## Summary by Sourcery

Fix the HTTP action schema to accept JSON arrays, validate the change with new tests, and prepare the 1.119.2 release

Bug Fixes:
- Allow the HTTP request schema to accept JSON arrays for the JSON field

Build:
- Bump HTTP module version to 1.119.2 in manifest.json

Documentation:
- Update CHANGELOG.md with the 1.119.2 release and fix entry

Tests:
- Add tests for POST requests with both JSON objects and JSON arrays

### DIFF
--- a/HTTP/action_request.json
+++ b/HTTP/action_request.json
@@ -29,7 +29,13 @@
       },
       "json": {
         "description": "The JSON to attach as body of the request",
-        "type": "object"
+        "oneOf": [
+          {"type": "object"},
+          {
+            "type": "array",
+            "items": {"type": "object"}
+          }
+        ]
       },
       "params": {
         "description": "Query string parameters to append to the URL",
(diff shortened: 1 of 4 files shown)


HTTP/action_request.json
```diff
@@ -29,7 +29,13 @@
...
       "json": {
         "description": "The JSON to attach as body of the request",
-        "type": "object"
+        "oneOf": [
+          {"type": "object"},
+          {
+            "type": "array",
+            "items": {"type": "object"}
+          }
+        ]
```
Darkheir, line 38: I think you can simply use `"type": ["object", "array"]`: https://json-schema.org/understanding-json-schema/reference/type#type-specific-keywords

Darkheir (approved):

HTTP/action_request.json
goudyj: I didn't know thanks!
However this current solution specifies the item type in the array. It doesn't allow having list of strings for instance, so I'm going to stay with that.

PR #1409, opened 2025-06-30
fix(HTTP): rollback to the previous version of request url

goudyj:
Rollback to the previous version since a front bug was introduced.


squioc (approved):
LGTM

PR #1411, opened 2025-07-01
Use warning level when unable to refresh stream

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/725

## Summary by Sourcery

Bug Fixes:
- Use warning level for logger and internal log when unable to refresh the event stream.


squioc (changes requested):
Can you update the version number in the manifest and add an entry in the changelog? Thank you

squioc (approved):
LGTM. Thank you for the work

PR #1413, opened 2025-07-02
Fix: Watchguard EPDR change default timediff to 7 days when filtering events

vg-svitla:
… events

## Summary by Sourcery

Update the WatchGuard EPDR connector to use a 7-day default event window instead of 1 hour, bump the version to 1.1.2, and record the change in the changelog.

Bug Fixes:
- Default to retrieving events from the last 7 days instead of 1 hour when no prior event timestamp is cached
- Use a 7-day lower bound when computing the max event date for retrieval

Build:
- Bump WatchGuard connector version to 1.1.2 in manifest.json

Documentation:
- Add 1.1.2 entry to CHANGELOG.md with details of the default timeframe fix

### DIFF
--- a/WatchGuard/watchguard/client/http_client.py
+++ b/WatchGuard/watchguard/client/http_client.py
@@ -7,7 +7,7 @@ from aiolimiter import AsyncLimiter
 from loguru import logger
 from pydantic.v1 import BaseModel
 
-from watchguard.client.errors import WatchGuardAuthError, WatchGuardError
+from watchguard.client.errors import WatchGuardError
 from watchguard.client.security_event import SecurityEvent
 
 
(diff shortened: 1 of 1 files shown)


WatchGuard/watchguard/epdr_connector.py
```diff
@@ -77,20 +77,20 @@ def last_event_date(self, security_event: SecurityEvent) -> datetime:
             datetime:
         """
         now = datetime.now(timezone.utc)
-        one_hour_ago = (now - timedelta(hours=1)).replace(microsecond=0)
+        one_day_ago = (now - timedelta(days=7)).replace(microsecond=0)
```
TOUFIKIzakarya, line 80: I think that here we can either change the variable name or days number

TOUFIKIzakarya (commented):
Thanks for the change.
Small change to do.

squioc (approved):
LGTM

PR #1420, opened 2025-07-11
fix(Triage) better handling of multiples files and signatures states

r1chev:
Following recent triage analysis we found out that the handling of multiple files with signatures or without any was not correct.
Trying to improve that with this PR.

## Summary by Sourcery

Improve signature checking in trigger_triage and update tests to cover multi-file and single-signature scenarios

Bug Fixes:
- Fix check_sample_signature to aggregate signature validity across all files and return true if any trusted signature is found

Enhancements:
- Initialize a default signature state and use any() to simplify the final signature check

Tests:
- Add a new static report fixture for a sample with a single signature
- Add a unit test to verify single-signature detection
- Update test mocks and comments to reflect multi-signature handling


Triage/triage_modules/trigger_triage.py
```diff
@@ -125,18 +125,21 @@ def get_malware_samples(self, malware: str, period: int) -> list[str]:
...
-            signers = static_report.get("files")[0]["metadata"]["pe"]["code_sign"]["signers"]
-            for value in signers:
-                if isinstance(value["validity"]["trusted"], list):
-                    for entry in value["validity"]["trusted"]:
-                        if entry:
-                            signature = entry
-                else:
-                    signature = value["validity"]["trusted"]
-            return signature
+            for f in static_report.get("files"):
```
squioc, line 130: NoneType is not iterable.
```suggestion
            for f in static_report.get("files", []):
```

squioc (changes requested):
Thank you for the job.
A little change to apply.

PR #1436, opened 2025-08-01
Harfanglab: fix checkpoint

TOUFIKIzakarya:
Fix checkpoint to send all asset for in the first run

### DIFF
--- a/HarfangLab/harfanglab/asset_connector/device_assets.py
+++ b/HarfangLab/harfanglab/asset_connector/device_assets.py
@@ -36,7 +36,6 @@ class HarfanglabAssetConnector(AssetConnector):
         super().__init__(*args, **kwargs)
         self.context = PersistentJSON("context.json", self._data_path)
 
-
     @property
     def most_recent_date_seen(self) -> datetime | None:
         with self.context as cache:
@@ -163,7 +162,8 @@ class HarfanglabAssetConnector(AssetConnector):
 
             if next_page := devices.get("next"):
                 next_page_url = urljoin(self.base_url, next_page)
-            else: return
+            else:
+                return
 
             device_response = self.client.get(next_page_url, params=params)
             device_response.raise_for_status()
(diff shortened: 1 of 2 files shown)


HarfangLab/harfanglab/asset_connector/device_assets.py
```diff
@@ -129,18 +132,18 @@ def map_fields(self, asset: dict[str, Any]) -> DeviceOCSFModel:
...
 
         params: dict[str, Union[str, int]] = {
             "ordering": self.DEVICE_ORDERING_FIELD,
-            "firstseen": firstseen_param,
             "limit": self.limit,
-            "offset": offset,
         }
 
+        if firstseen_param:
+            params["firstseen"] = firstseen_param
```
squioc, line 145: Proposal
```suggestion
        if from_date:
            params["firstseen"] = from_date
```
And we can remove L137

HarfangLab/harfanglab/asset_connector/device_assets.py
```diff
@@ -154,37 +157,35 @@ def __fetch_devices(self, from_date: datetime) -> Generator[list[dict[str, Any]]
...
+                next_page_url = urljoin(self.base_url, next_page)
+            else:
                 return
 
-            # Update the offset for the next request
-            offset += self.limit
-            params["offset"] = offset
-
-            device_response = self.client.get(next_page, params=params)
+            device_response = self.client.get(next_page_url, params=params)
```
squioc, line 165: If the `from_date` argument is not none, `params` will contain `firstseen` in the list of parameters.
Is something expected? Won't the `next` property, in the response, set some parameters?

squioc (approved):
LGTM.
Some questions were asked

HarfangLab/harfanglab/asset_connector/device_assets.py
TOUFIKIzakarya: Yes you are right @squioc, good catch.

squioc (approved):
LGTM

PR #1443, opened 2025-08-05
HornetSecurity: fix the batch system

squioc:
- Move the computation of the events collect duration in `fetch_events` as the main loop stays in the TimeStepper (#L106)
- Move the accumulative batch of events in the `self.fetch_events()` loop, for small batches (Otherwise, the collector wait reaching the threshold to send the events).

## Summary by Sourcery

Refine the event batching mechanism by computing fetch durations in the main fetch loop, sending small batches immediately, updating related metrics/logs, and bumping the connector version

Enhancements:
- Move duration measurement and event count logging into the fetch_events time-step loop
- Accumulate and forward small batches inside the fetch_events loop instead of waiting for next_batch to finish
- Update Prometheus metrics for fetch and forward durations per batch

Build:
- Bump HornetSecurity connector version to 1.0.3

Documentation:
- Add changelog entry for batch system fix in version 1.0.3

### DIFF
--- a/HornetSecurity/CHANGELOG.md
+++ b/HornetSecurity/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix the batch system for little batches
+- Fix the batch system for small batches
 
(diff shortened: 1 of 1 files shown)


mchupeau-sk (approved):
LGFM

PR #1452, opened 2025-08-11
TrendMicro VisionOne - rename triggers and fix actions

lvoloshyn-sekoia:
1. TrendMicro VisionOne API allows batch requests. So for the Terminate Process action we tried to implement `endpoint id`(agent GUID) and `process id` as list of objects (because process id is different for each endpoint). Frontend doesn't support this, so we decided to replace a batch action with a single-endpoint action, thus using `agent guid` and `process id` as single keyword fields.

2. Original Trend Micro module has unclear labels. I added fix for this as well
image

## Summary by Sourcery

Adapt the TrendMicro VisionOne Terminate Process action to accept a single agent and process pair instead of a batch list, clarify module labels, update tests, and bump version numbers with an updated changelog.

Enhancements:
- Refactor Terminate Process action to use individual agent_guid and process_id fields instead of a list of objects
- Clarify ambiguous labels in the core TrendMicro module
- Bump connector and module versions in manifest files and add a 1.6.2 release entry in the CHANGELOG

Documentation:
- Add VisionOne separation note and new version entry to CHANGELOG

Tests:
- Update terminate process action tests to match the new argument structure

### DIFF
--- a/TrendMicroVisionOne/CHANGELOG.md
+++ b/TrendMicroVisionOne/CHANGELOG.md
@@ -6,3 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## 2025-08-08 - 0.1.1
+
+### Fixed
+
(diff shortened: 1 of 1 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #1459, opened 2025-08-14
VECTRA : add user agent in client

mchupeau-sk:
https://github.com/SekoiaLab/integration/issues/771

## Summary by Sourcery

New Features:
- Include a User-Agent header with the connector version in the client headers


squioc (changes requested):
Thank you for the work.
@mchupeau-sk Can you bump the version number in the manifest file and add an entry in the CHANGELOG?

PR #1473, opened 2025-08-28
SentinelOne: force hostname parameter to be without scheme

squioc:
Fix the module configuration to force the hostname to be without https://


LenaigKaliou (approved):
LGTM

PR #1502, opened 2025-09-18
fix error verticles collector init

mchupeau-sk:
https://github.com/SekoiaLab/integration/issues/807

## Summary by Sourcery

Ensure verticles_collector property handles get_edge_types errors gracefully by instantiating the collector inside the try block and streamline tests to focus on client.get_edge_types failures

Bug Fixes:
- Delay VerticlesCollector instantiation to inside the try block in the verticles_collector property so that client errors are caught before initialization

Enhancements:
- Simplify verticles_collector tests by patching trigger.client.get_edge_types directly instead of mocking the VerticlesCollector class

Tests:
- Update tests for 403 errors, other HTTP errors, and general exceptions to use patch.object on trigger.client.get_edge_types with appropriate side effects


squioc (approved):
LGTM. Thank you for the fix

PR #1519, opened 2025-09-26
Fix: CrowdStrike OOM Issue (921)

vg-svitla:
Relates to [921](https://github.com/SekoiaLab/integration/issues/921)


squioc (changes requested):
Thank you for the modifications.

In the good direction. However, we have some changes to apply

PR #1526, opened 2025-10-01
Hotfix: Delete Case Action (937)

vg-svitla:
Relates to [937](https://github.com/SekoiaLab/integration/issues/937)
More detailed possible fix is in sdk

## Summary by Sourcery

Bug Fixes:
- Rename DeleteCase registration route parameter from uuid to case_uuid


squioc (changes requested):
Hello,
Thank you for the fix.

Can you add an entry to the changelog and bump the version number in the manifest?

PR #1529, opened 2025-10-02
[SEKOIA.IO] Add retry strategy 

mchupeau-sk:
https://github.com/SekoiaLab/integration/issues/956


squioc (changes requested):

squioc (changes requested):
LGTM.
Can you bump the version number in the manifest and add an entry in the changelog?

PR #1531, opened 2025-10-03
HTTP: Improve Request URL action

squioc:
Allow to supply a dictionary in the `params` arguments.
Currently, only a string was allowed


mchupeau-sk (approved):
LGFM

PR #1533, opened 2025-10-06
Sekoia.io: Renew Get Asset V2 action

squioc:
- Deprecate the old `Get Asset V2` that map the result according to the AssetManagementV1 API response
- Add new `Get Asset V2` that return the response of the AssetManagementV2 API directly


mchupeau-sk (approved):
LGFM

PR #1605, opened 2025-12-01
fix okta

mchupeau-sk:
## Summary by Sourcery

Bug Fixes:
- Fix Okta client configuration by passing the API key directly instead of using SecretStr accessors that caused incompatibilities.


TOUFIKIzakarya (approved):
LGFM

PR #1664, opened 2025-12-09
Ubika : Add timestepper to connector

TOUFIKIzakarya:
Add timestepper to ubika
Related issue : [issue](https://github.com/SekoiaLab/integration/issues/1292) 

## Summary by Sourcery

Introduce a time-stepped polling mechanism for Ubika Cloud Protector Next Gen connector and persist the last processed timestamp to control event retrieval windows.

New Features:
- Add a configurable TimeStepper utility to generate time ranges for event collection based on frequency, timedelta, and start time.
- Persist and reuse the most recent processed event time in connector context to resume polling from the correct point across runs.

Bug Fixes:
- Avoid failures when AuthorizationError is raised without a secondary message argument by safely logging the error.

Enhancements:
- Refactor the connector to fetch events between explicit start and end timestamps and drive batching via the new TimeStepper ranges.
- Cap the historical resume point to at most one month in the past to limit backfill range and control lag metrics reporting.

Tests:
- Adapt existing connector tests to the new time-ranged fetching API and implement comprehensive unit tests for the TimeStepper behavior, including lag logging and sleep handling.


Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen.py
```diff
@@ -161,42 +206,17 @@ def __fetch_next_events(self, from_timestamp: int) -> Generator[list, None, None
...
-            # save the most recent date
-            if most_recent_date_seen > self.from_timestamp:
-                self.from_timestamp = most_recent_date_seen
-
-                # save in context the most recent date seen
-                self.cursor.offset = self.from_timestamp
-
-                current_lag = time.time() - most_recent_date_seen
-
-        EVENTS_LAG.labels(intake_key=self.configuration.intake_key).set(current_lag)
```
squioc: With this removal, we are losing the `EVENTS_LAG`, can you reintroduce it?

Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen.py
```diff
@@ -225,22 +245,20 @@ def next_batch(self) -> None:
...
+                break
+
             try:
-                self.next_batch()
+                self.next_batch(start, end)
             except Exception as error:
                 self.log_exception(error, message="Failed to forward events")  # pragma: no cover
+            finally:
+                with self.context as cache:
+                    cache["most_recent_date_seen"] = end.isoformat()
```
squioc: I suggest moving this code with the `self.next_batch` method and to move loop in the try-catch.
Currently, if we faces an issue with the API, we will continue to progress and lost events.

squioc (changes requested):
Thank you for the fixes.
Some changes to apply

Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen.py
TOUFIKIzakarya: Yes thanks for this catch.

TOUFIKIzakarya: I added that but i notice that we have the lag in the timestepper script

squioc (approved):
LGTM. Thank you for the work.

PR #1666, opened 2025-12-10
Sophos XDR connector - fix cursor

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/1176

## Summary by Sourcery

Fix Sophos XDR query trigger cursor handling by making the query definition a property and ensuring it is always derived from the latest from_date.

Bug Fixes:
- Ensure the Sophos XDR IOC query uses a dynamic query property so the from cursor is correctly updated based on the latest from_date.

Documentation:
- Correct minor typos in the Sophos XDR query trigger class docstring.

Tests:
- Extend Sophos XDR query tests to validate that the query from field matches the current from_date cursor.

### DIFF
--- a/Sophos/manifest.json
+++ b/Sophos/manifest.json
@@ -10,7 +10,7 @@
         "format": "uri"
       },
       "api_host": {
-        "description": "API Url of the Sophos Central API (e.g. 'https://api-{dataRegion}.central.sophos.com')",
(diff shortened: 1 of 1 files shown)


TOUFIKIzakarya (commented):
Thanks for the fix.
Can you please update the changelog and manifest.

PR #1672, opened 2025-12-11
Fix: Azure Event Hub Resource Not Found (1167)

vg-svitla:
Relates to [1167](https://github.com/SekoiaLab/integration/issues/1167)

## Summary by Sourcery

Address Azure Event Hub resource-not-found issues by improving checkpoint store initialization and handling missing resources more gracefully.

Bug Fixes:
- Pre-create the Azure Blob Storage container used for checkpoints to avoid 404 errors during partition ownership claims.
- Handle Event Hub ResourceNotFound errors during event reception by logging a warning instead of failing the connector.

Build:
- Bump Azure integration manifest version from 2.9.1 to 2.9.2.

Documentation:
- Update Azure connector changelog with a 2.9.2 release entry describing the resource-not-found fix and checkpoint store initialization.

### DIFF
--- a/Azure/connectors/azure_eventhub.py
+++ b/Azure/connectors/azure_eventhub.py
@@ -82,25 +82,27 @@ class AzureEventsHubTrigger(AsyncConnector):
     def client(self) -> Client:
         return Client(self.configuration)
 
-    async def _initialize_checkpoint_store(self) -> None:
+    async def _initialize_checkpoint_store(self) -> None:  # pragma: no cover
         """
         Pre-create the storage container to prevent race conditions during partition ownership claims.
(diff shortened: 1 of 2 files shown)


Azure/connectors/azure_eventhub.py
```diff
@@ -80,6 +82,27 @@ def __init__(self, *args: Any, **kwargs: Optional[Any]) -> None:
     def client(self) -> Client:
         return Client(self.configuration)
 
+    async def _initialize_checkpoint_store(self) -> None:
```
TOUFIKIzakarya: Can't see where this function is used ? maybe we should use it on the async_run() ?

TOUFIKIzakarya (commented):
Thanks for the fix.
Small question

PR #1694, opened 2025-12-16
fixing invalid json requests following TLP and PAP updates

CharlesLR-sekoia:
## Summary by Sourcery

Normalize TLP and PAP values and improve observable generation for TheHive connector.

Bug Fixes:
- Convert string TLP and PAP values to their corresponding numeric levels before sending observables to avoid invalid JSON requests.
- Skip observables with unknown data types instead of emitting invalid payloads.

Enhancements:
- Ensure observable data is always serialized as a string when building TheHive observables.
- Log warnings for unknown TLP and PAP levels and for observables with unknown data types to aid troubleshooting.

### DIFF
--- a/TheHiveV5/CHANGELOG.md
+++ b/TheHiveV5/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


squioc (changes requested):
Thanks for the fixes.
Can you increase the version number in the manifest and add an entry in the README.md :pray: ?

PR #1845, opened 2026-01-08
Cybereason: Reintroduce the one millisecond

squioc:
- Add back the one millisecond to exclude already collected events from the next batch

## Summary by Sourcery

Prevent duplicate event collection by adjusting the stored timestamp between batches and release a new Cybereason connector patch version.

Bug Fixes:
- Avoid re-fetching the last processed events by advancing the most recent event timestamp by one millisecond when updating from_date.

Build:
- Bump Cybereason connector manifest version from 1.13.4 to 1.13.5.

Documentation:
- Document the fix and new 1.13.5 release in the Cybereason changelog.

Tests:
- Update connector pull events tests to assert the adjusted timestamp behavior for subsequent batches.

### DIFF
--- a/Cybereason/CHANGELOG.md
+++ b/Cybereason/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-01-08 - 1.13.5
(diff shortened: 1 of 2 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #1957, opened 2026-01-27
HarfangLab - fix actions

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/1252

## Summary by Sourcery

Refactor HarfangLab job execution to use a shared API client and return actual process/pipe results, while updating related actions, tests, and configuration.

New Features:
- Allow Get Processes and Get Pipes actions to return collected items directly or save them to a file.

Bug Fixes:
- Ensure job-related actions wait for job completion and correctly fetch paginated results from HarfangLab endpoints.
- Fix download-file workflow to use explicit job IDs and artefact metadata instead of shared executor state.
- Normalize API base URL handling in the client to avoid trailing-slash issues.
- Make hostname lookup use the shared API client for authenticated requests.

Enhancements:
- Introduce a reusable JobExecutor client abstraction with pagination and file-saving helpers.
- Simplify artefact download logic by passing explicit artefact info and removing hidden state.
- Update tests to exercise real HTTP flow via requests-mock instead of internal executor state mocking.

Documentation:
- Update changelog with version 1.29.0 and description of the changed job flow and new action outputs.

Tests:
- Add and update tests for process/pipe listing and file download actions to cover the new job flow and result formats, and configure pytest defaults and coverage thresholds.

Chores:
- Switch models to use pydantic v1 import for compatibility.
- Bump integration manifest version to 1.29.0.

### DIFF
--- a/HarfangLab/action_harfanglab_getpipelist.json
+++ b/HarfangLab/action_harfanglab_getpipelist.json
@@ -15,7 +15,7 @@
     "title": "Arguments",
     "type": "object"
   },
-  "description": "Get the list of named pipe on the systems",
+  "description": "Get the list of named pipe on the systems (deprecated)",
   "docker_parameters": "harfanglab_get_pipe_list",
   "name": "List named pipes",
   "results": {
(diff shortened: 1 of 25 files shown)


HarfangLab/action_harfanglab_getpipelist.json
squioc: Modifying the output properties wil break existing playbooks (even if the action doesn't work well).
Can you duplicate the action (JSON manifest file and Python code) and deprecated the old version?

PR #2066, opened 2026-02-26
fix(http): Wrong path for download_file action

CharlesLR-sekoia:
## Summary by Sourcery

Return both absolute and data-path-relative download paths from the HTTP download_file action and update versioning metadata.

Bug Fixes:
- Ensure the download_file action exposes a correct relative path under the configured data directory while still returning the absolute file path.

Documentation:
- Document the relative path fix in the HTTP module changelog.

Tests:
- Extend download_file tests to validate absolute vs relative paths and their consistency.

Chores:
- Bump HTTP module version to 1.120.2 in the manifest.


TOUFIKIzakarya (approved):
LGFM

PR #2072, opened 2026-03-04
Imp/disable user from email

mchupeau-sk:
https://github.com/SekoiaLab/integration/issues/1233

## Summary by Sourcery

Switch user search and account actions to filter by email address instead of display name, and tighten validation of the search inputs.

Bug Fixes:
- Ensure user search raises an error when neither username nor email is provided, preventing ambiguous or invalid queries.

Enhancements:
- Support searching users by email alone or in combination with username, updating filters to use the mail attribute.
- Update user account action arguments to accept an optional email filter and propagate it through disable, enable, and reset-password flows.
- Adjust search filter construction to avoid unnecessary composite filters when only username is provided.

Tests:
- Update and extend tests to cover email-based search filters, username-only searches, email-only searches, and input validation scenarios.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,4 @@ dmypy.json
 .vscode/
 */.devcontainer/*
 */test_end_to_end.py
+MicrosoftActiveDirectory/test_end_to_end_action.py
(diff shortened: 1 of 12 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #2129, opened 2026-03-13
Withsecure: fix payload

TOUFIKIzakarya:
Main issue : [issue](https://github.com/SekoiaLab/integration/issues/1383)

## Summary by Sourcery

Ensure WithSecure security event connector sends engineGroup filters correctly using form-urlencoded requests and is covered by tests.

Bug Fixes:
- Fix security events API request payload to repeat engineGroup parameters in application/x-www-form-urlencoded format instead of sending a list structure.

Tests:
- Add integration-style test to verify engineGroup parameters are repeated in the request body and that the correct form-urlencoded Content-Type header is used when fetching events.

### DIFF
--- a/WithSecure/CHANGELOG.md
+++ b/WithSecure/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


squioc (changes requested):
LGTM. Thank you for the fix

PR #2147, opened 2026-03-19
Fix: Aws Additional Filters (1431)

vg-svitla:
Relates to [1431](https://github.com/SekoiaLab/integration/issues/1431)

## Summary by Sourcery

Add optional S3 object key prefix filtering to SQS-based S3 queued connectors and bump the AWS integration version.

New Features:
- Introduce an optional prefix_filter configuration on S3 queued connectors to only process S3 objects whose keys start with a specified prefix.

Enhancements:
- Skip non-matching S3 objects based on the configured key prefix while logging debug information about skipped objects.

Build:
- Update AWS integration manifest version to 1.34.0.

Documentation:
- Document the new prefix_filter option and its use case in the AWS connector changelog.

Tests:
- Add tests verifying that S3 queued connectors process only messages whose object keys match the configured prefix and that all messages are processed when no prefix_filter is set.

### DIFF
--- a/AWS/CHANGELOG.md
+++ b/AWS/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-03-17 - 1.34.0
+
+### Added
+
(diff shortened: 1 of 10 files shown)


mchupeau-sk (approved):
LGFM

PR #2416, opened 2026-04-13
Feat: Add ilert module

DaTiMy:
## Summary by Sourcery

Add a new ilert automation module that forwards Sekoia.io alerts to the ilert Events API as a trigger alert action.

New Features:
- Introduce an ilert module with a Trigger Alert action that fetches a Sekoia.io alert and posts it to the ilert Events API using a configurable integration key and endpoint.
- Add retry-enabled HTTP helper utilities for resilient communication with the ilert Events API.
- Expose module configuration via a manifest defining ilert integration settings.

Build:
- Add Poetry-based project configuration and a Dockerfile to build and run the ilert automation module container.

Documentation:
- Add an initial changelog describing the first ilert module release.

Tests:
- Add unit tests covering the ilert trigger alert action behavior and the HTTP retry helper utility.


mchupeau-sk:
Hello @DaTiMy !

First review pass found a few blocking issues in this PR:

1. The happy-path test is currently failing.
   `IlertTriggerAlertAction.run()` adds a mapped `status` field to the payload before posting to ilert, but `test_ilert_postalert_default` still asserts that the POST body is exactly the original `alert_info` dict. I ran `poetry run pytest tests -q` in `ilert/` and got `1 failed, 1 passed`, with this assertion as the failure.

2. `base_url` handling is brittle and can generate an invalid Sekoia URL.
   `_get_alert()` builds the URL with direct string concatenation: `f"{base_url}v1/sic/alerts/{alert_uuid}"`. If the caller provides `https://api.sekoia.io` instead of `https://api.sekoia.io/`, the request becomes `https://api.sekoia.iov1/...`. The action JSON only validates `base_url` as a generic URI, so this is a real runtime footgun.

3. Retry logic is only applied to the ilert POST, not to the initial Sekoia GET.
   The outbound ilert request uses `requests_retry_session()`, but `_get_alert()` uses plain `requests.get()`. A transient 429/5xx from Sekoia will fail the whole action immediately even though the module already includes retry infrastructure for the same class of failures.

Non-blocking but worth tightening before this grows further:

- Add a test for `base_url` without a trailing slash.
- Add coverage for the `_map_status()` branches (`closed`, `resolved`, `acknowledged`, malformed values).
- Strengthen `test_helpers.py`: right now it only checks that a session object exists, not that retries are configured or effective.

DaTiMy:
@mchupeau-sk Thank you for the review! I've pushed the changes

mchupeau-sk (approved):
Thanks for the work ! It will be available soon on the production platform

mchupeau-sk:
Hello @DaTiMy

To continue the homologation process can you create a PR in the documentation [here](https://github.com/SEKOIA-IO/documentation/tree/main/docs/integration/).

NB : for the shared part, it's generated automatically.

Thanks in advance !

PR #2506, opened 2026-04-23
fix(sophos): fix sophos whoami url

TOUFIKIzakarya:
As i have 504 http error on sophos asset connector. So i changed the whoami url.
I don't know why is it working for other event connectors

## Summary by Sourcery

Update the Sophos connector to use the documented global whoami endpoint and bump the connector version.

Bug Fixes:
- Fix Sophos Device Asset Connector whoami calls by switching to the official global whoami URL.

Documentation:
- Add changelog entry documenting the whoami URL change for the Sophos Device Asset Connector.

### DIFF
--- a/Sophos/sophos_module/client/auth.py
+++ b/Sophos/sophos_module/client/auth.py
@@ -87,8 +87,9 @@ class SophosApiAuthentication(AuthBase):
         Call the Whoami endpoint to get tenancy information
         """
         # Get the whoami
+        # Doc: https://developer.sophos.com/intro
         response = self.__http_session.get(
-            url=f"{self.__api_host}/whoami/v1",
+            url="https://api.central.sophos.com/whoami/v1",
             headers={"Authorization": credentials.authorization},
         )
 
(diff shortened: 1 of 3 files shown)


Sophos/sophos_module/client/auth.py
```diff
@@ -87,8 +87,9 @@ def __whoami(self, credentials: SophosApiCredentials) -> Any:
         Call the Whoami endpoint to get tenancy information
         """
         # Get the whoami
+        # Doc: https://developer.sophos.com/intro
         response = self.__http_session.get(
-            url=f"{self.__api_host}/whoami/v1",
+            url="https://api.central.sophos.com/whoami/v1",
```
squioc, line 92: The base url is supplied through the configuration because Sophos have regions and the whoami API is available in each API.
Using the golbal url can break existing integration

squioc (changes requested):
Thank you for the proposal.

Before merging, we have to check if it don't break exisintg integration.

TOUFIKIzakarya:
Hello,

I will close this PR, as we provide this URL in the arguments and we can do it without region.

PR #2509, opened 2026-04-24
Fix: MicrosoftDefender New Connector (1495)

vg-svitla:
Relates to [1495](https://github.com/SekoiaLab/integration/issues/1495)


squioc (changes requested):
Hello,

Thank you for the proposal. However, to be consistent with the rest of the automation mation, Can you use Microsoft GraphAPI to collect the incidents (see [doc](https://learn.microsoft.com/en-us/graph/api/security-list-incidents?view=graph-rest-1.0&tabs=http)?
Thank you.

MicrosoftDefender/microsoftdefender_modules/connector_microsoft_defender_xdr.py
squioc, line 1: It would be greate having the base classes in a dedicated file. It will simplify the alert connector also.

MicrosoftDefender/microsoftdefender_modules/connector_defender_incidents.py
```diff
@@ -0,0 +1,32 @@
...
+
+    endpoint_url = "https://graph.microsoft.com/v1.0/security/incidents"
+    timestamp_field = "createdDateTime"
+    id_field = "id"
+    context_cursor_key = "most_recent_date_requested_incidents"
+    events_cache_context_key = "incidents_events_cache"
+
+    def extra_query_params(self) -> dict[str, Any]:
+        if self.configuration.expand_alerts:
+            return {"$expand": "alerts"}
```
squioc: I wonder if this parameter is useful. According to the intake format, we only count the number of alerts associated to the incident:
1. this number can evolve with the time
2. retrieve a list of alerts, from the API, to count them seems to be overkilled.

I suggest removing this parameter from this code and from the configuration files.

squioc (changes requested):
Thank you @vg-svitla  for the work.
I put some suggestions.

MicrosoftDefender/microsoftdefender_modules/connector_defender_incidents.py
vg-svitla: Done

MicrosoftDefender/microsoftdefender_modules/connector_microsoft_defender_xdr.py
vg-svitla: Done

TOUFIKIzakarya (approved):
LGFM.
Just for the changelog we have 2 versions, can we merge them in one version ?

PR #2532, opened 2026-04-29
Set TLS certificate verification status to False in FortigateDisableLocalUserAction to make it work with firewalls configured with the default fortigate self-signed certificate

kossithedon:
Set TLS certificate verification status to False in FortigateDisableLocalUserAction to make it work with firewalls configured with the default fortigate self-signed certificate

## Summary by Sourcery

Support FortiGate firewalls with self-signed certificates by making TLS certificate verification configurable.

New Features:
- Add a configurable FortiGate certificate verification option, enabled by default.

Bug Fixes:
- Allow FortiGate actions to work with firewalls using the default self-signed certificate when certificate verification is disabled.

Chores:
- Refresh the FortiGate dependency lockfile.


kossithedon:
The suggested modifs should not be applied in order to allows to use this action with firewalls with self-signed certificates.

squioc (changes requested):
Hello @kossithedon,

Thank you for your proposal.
After reviewing the changes, we think that our proposal should be mutualized and extended to other Fortigate actions with the ability to set the TLS certificates verification from the configuration.
Can you add a new parameters in the properites of the manifest file (like [TheHiveV5](https://github.com/SEKOIA-IO/automation-library/blob/develop/TheHiveV5/manifest.json#L18-L22)), set the verify argument from the module configuration (see [The Hive](https://github.com/SEKOIA-IO/automation-library/blob/develop/TheHiveV5/thehive/upload_logs.py#L17)) and extended it to all actions?

In addition, can you bump the version number in the manifest file and add an entry in the CHANGELOG.md?
Thank you

kossithedon:
Hi,

Sure ! it's a better idea.

I just made the changes.

kossithedon:
Hi,

Just a chase up here

kossithedon:
Hi,

I just merged modifications proposed by @tfanouillere-sekoia. Can you check and accept the changes ?

PR #2541, opened 2026-05-04
fix(sophos): fix samples

TOUFIKIzakarya:
## Summary by Sourcery

Bump the Sophos integration to version 1.18.2 and correct the embedded JSON samples in the asset connector mapping configuration.

Bug Fixes:
- Fix improperly formatted sample JSON blocks in the asset connector device mapping configuration so they are treated as literal data instead of YAML structures.

Documentation:
- Add a changelog entry documenting the 1.18.2 release and the fix to the asset connector samples.


mchupeau-sk (approved):
LGFM

PR #2592, opened 2026-05-15
Fix/1587 crowdstrike falcon

mchupeau-sk:
https://github.com/SekoiaLab/integration/issues/1587

## Summary by Sourcery

Bug Fixes:
- Avoid decreasing cached stream offsets when processing batches that contain a lower offset than one previously stored.


squioc (approved):
LGTM. Thank you for the work.

Maybe a test for the change would be great.

PR #2597, opened 2026-05-19
CrowdStrikeFalcon: stop the connectors on authentication error

squioc:
- Refactor the authentication module to raise a specific exception for 401 and 403 HTTP errors
- Log and stop the connectors when facing an AuthenticationError exception

## Summary by Sourcery

Handle CrowdStrike Falcon authentication failures explicitly and ensure connectors stop gracefully on authentication errors.

Enhancements:
- Introduce a dedicated AuthenticationError exception for 401/403 responses from the CrowdStrike Falcon token endpoint and use it in the authentication flow.
- Update event stream triggers and asset connectors to catch authentication errors, log them as critical, and stop processing instead of continuing.
- Align asset connector tests with the real module configuration and add coverage for authentication error handling in both device and user asset connectors, as well as the authentication client.

Documentation:
- Document the new behaviour of stopping connectors on authentication errors in the changelog.

### DIFF
--- a/CrowdStrikeFalcon/CHANGELOG.md
+++ b/CrowdStrikeFalcon/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Stop the connector on authentication error
+- Stop the connector when authentication fails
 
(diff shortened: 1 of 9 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #2691, opened 2026-06-02
fix filter request

mchupeau-sk:
## Summary by Sourcery

Fix device asset checkpoint handling to use a valid, URL-safe last-seen timestamp filter and update stored checkpoint format.

Bug Fixes:
- Use the Defender machine lastSeen field instead of firstSeen for incremental checkpoint filtering in the device asset connector.
- Store checkpoints as UTC timestamps with millisecond precision and Z suffix to avoid URL query string issues when building filter parameters.

Build:
- Bump the Microsoft Defender integration manifest version to 1.2.2.

Documentation:
- Document the device asset connector checkpoint filter fix and timestamp encoding details in the changelog.

Tests:
- Update device asset connector tests to validate URL-encoded lastSeen filter usage, absence of filters without checkpoints, and the new checkpoint timestamp format.

### DIFF
--- a/MicrosoftDefender/CHANGELOG.md
+++ b/MicrosoftDefender/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## 2026-06-02 - 1.2.2
+
+### Fixed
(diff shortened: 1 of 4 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #2542, opened 2026-05-04
feature/ubika_new_intake_traffic_logs

clement-burtscher-sekoia:
## Description

Fixes https://github.com/SekoiaLab/detection-rules/issues/3877

### Added

- Add new `connector_ubika_cloud_protector_next_gen_traffic_logs` connector

### Changed

- Refactor `next_gen` connectors and reduce code duplication by creating the `connector_ubika_cloud_protector_next_gen_base` connector
- Rename `connector_ubika_cloud_protector_next_gen` -> `connector_ubika_cloud_protector_next_gen_alerts`

PR #2694, opened 2026-06-03
fix/ubika_cloud_protector_next_gen_client_closed_error

clement-burtscher-sekoia:
## Description

Fixes https://github.com/SekoiaLab/integration/issues/1631

### Fixed

- Recreate the cached NextGen HTTP client after shutdown to prevent closed-client request failures on connector restart

## Summary by Sourcery

Ensure the Ubika Cloud Protector NextGen connector recreates its HTTP client after shutdown to avoid using a closed client across runs.

Bug Fixes:
- Invalidate the cached NextGen HTTP client after each run so subsequent runs create a fresh client instead of reusing a closed one, preventing closed-client request failures on connector restarts.

Build:
- Bump the Ubika integration manifest version from 1.1.0 to 1.1.1.

Documentation:
- Update the changelog with a 1.1.1 release entry describing the HTTP client shutdown fix.

Tests:
- Add tests verifying that the connector run invalidates the cached client both on exceptions and on clean shutdowns.

### DIFF
--- a/Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen_base.py
+++ b/Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen_base.py
@@ -358,10 +358,9 @@ class UbikaCloudProtectorNextGenBaseConnector(Connector):
 
         finally:
             # Cleanup on stop or fatal error
-            self.client.close()
-            # Invalidate the cached client so a fresh one is created on the next run,
-            # preventing "Cannot send a request, as the client has been closed." errors
-            # when the framework restarts run() on the same instance.
-            self.__dict__.pop("client", None)
+            client = self.__dict__.pop("client", None)
+            if client is not None:
+                # Close + invalidate cached client to avoid reusing a closed client across runs.
+                client.close()
             self.save_events_cache()
             self.log(message=f"Stopped fetching {self.NAME} events", level="info")
--- a/Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen_base.py
+++ b/Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen_base.py
@@ -359,5 +359,9 @@ class UbikaCloudProtectorNextGenBaseConnector(Connector):
         finally:
             # Cleanup on stop or fatal error
             self.client.close()
+            # Invalidate the cached client so a fresh one is created on the next run,
+            # preventing "Cannot send a request, as the client has been closed." errors
+            # when the framework restarts run() on the same instance.
+            self.__dict__.pop("client", None)
             self.save_events_cache()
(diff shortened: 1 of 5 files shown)


Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen_base.py
```diff
@@ -358,6 +358,9 @@ def run(self) -> None:
 
         finally:
             # Cleanup on stop or fatal error
-            self.client.close()
+            client = self.__dict__.pop("client", None)
```
vg-svitla: Personally I would propose to create a private prop like below
```
_client: UbikaCloudProtectorNextGenApiClient | None = None
```

Then cached property changed to 
```
def client(self) -> UbikaCloudProtectorNextGenApiClient:
        """
        HTTP client that automatically injects tokens and handles rate‐limits.
        """
        if self._client is None:
           self._client = UbikaCloudProtectorNextGenApiClient(refresh_token=self.configuration.refresh_token)

        return self._client
```

So you can use it without any pop etc... Bug again it just my vision. Other then that its fine for me but lets wait for other review

clement-burtscher-sekoia: For information, before merging the previous PR [feature/ubika_new_intake_traffic_logs](https://github.com/SEKOIA-IO/automation-library/pull/2542) the code was:

```python
    @cached_property
    def client(self) -> UbikaCloudProtectorNextGenApiClient:
        return UbikaCloudProtectorNextGenApiClient(refresh_token=self.configuration.refresh_token)

    def run(self) -> None:
        [...]
        # Close the client connection
        if "client" in self.__dict__:
            self.client.close()
```
https://github.com/SEKOIA-IO/automation-library/blob/2f7ee1aa2d7dc16b59c7b6a958e723b20bcf16cf/Ubika/ubika_modules/connector_ubika_cloud_protector_next_gen.py#L268-L270

mchupeau-sk: I'm ok with the proposition of @vg-svitla :grin: 

but for me we need also to remove the 

> @cached_property

from the client definition and replace it by 

`@property`

clement-burtscher-sekoia: @vg-svitla @mchupeau-sk Fixed with commit [Set lazy init of HTTP client](https://github.com/SEKOIA-IO/automation-library/pull/2694/changes/7f0c14cb73886edd06a81bcf0aec7ffd1f52d0bc)

mchupeau-sk (approved):
LGFM

PR #2718, opened 2026-06-04
Fix: Search Alerts Fix Params (86881)

vg-svitla:
Relates to [86881](https://github.com/SekoiaLab/platform/issues/86881)

## Summary by Sourcery

Fix operation-center search requests so only user-configured filters are sent to the API.

Bug Fixes:
- Prevent Search Alerts, Search Cases, and Get Alert actions from sending unset query parameters that were incorrectly interpreted as active filters.
- Normalize retained boolean query parameters to lowercase API-compatible values.

Enhancements:
- Introduce shared query-parameter filtering for operation-center API actions.

Tests:
- Add coverage confirming unset filters are omitted and selected values, including booleans and numeric zeroes, are preserved.

Chores:
- Bump the Sekoia.io module version to 2.76.1 and document the fix in the changelog.


mchupeau-sk:
@vg-svitla  can we simplify the solution using pydantic option exclude_none, I'm not sur is just a question :grin:

vg-svitla:
`GetAlert` already uses `exclude_none`, but the node sends `""` and `False` rather than `None`, so it can't reach this bug on its own. I kept it and put the `""/False` drop in a shared `FilteredQueryParametersAction` base, now used by `Search Alerts`, `Search Cases` and `Get Alert`

mchupeau-sk (approved):

PR #2783, opened 2026-06-16
prevent system log re-ingestion after connector restart

tfanouillere-sekoia:
https://github.com/SekoiaLab/integration/issues/1683

## Summary by Sourcery

Improve Okta System Log connector deduplication and checkpoint handling to avoid event re-ingestion across restarts while keeping at-least-once delivery.

Bug Fixes:
- Prevent re-ingestion of previously processed Okta System Log events after connector restarts by persisting and consulting a robust deduplication cache.
- Ensure checkpoints are only advanced after successful event push so failed pushes do not skip or lose events.

Enhancements:
- Introduce checksum-based deduplication as a fallback for Okta System Log events without stable UUIDs, using key event attributes.
- Persist and reuse both event UUIDs and content checksums in the events cache to make deduplication resilient across restarts.
- Refine checkpoint computation into dedicated helpers for clarity and more predictable offset handling.

Tests:
- Extend System Log connector tests to cover checkpoint updates on success/failure, checksum-based deduplication behavior, cache persistence across batches, and cache sizing.

### DIFF
--- a/Okta/okta_modules/asset_connector/user_assets.py
+++ b/Okta/okta_modules/asset_connector/user_assets.py
@@ -11,9 +11,9 @@ from typing import Any, Optional
 
 from dateutil.parser import isoparse
 from okta.client import Client as OktaClient
-from okta.models.user import User as OktaUser
 from okta.models.role import Role as OktaRole
 from okta.models.role_status import RoleStatus as OktaRoleStatus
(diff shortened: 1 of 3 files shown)


mchupeau-sk:
@tfanouillere-sekoia I close this pr, Zakarya found the problem it was just an issue on the naming file ...

PR #2789, opened 2026-06-16
fix(okta): Fix the deplucation of events

TOUFIKIzakarya:
Main issue : [Issue](https://github.com/SekoiaLab/integration/issues/1683)

## Summary by Sourcery

Prevent duplicated Okta system log events by separating the events deduplication cache from the checkpoint state and ensuring both persist across connector restarts.

Bug Fixes:
- Store the system log events deduplication cache in its own persistent file so it no longer overwrites the checkpoint state on disk.
- Add backward-compatible loading of the events cache from the legacy context file to preserve existing deployments.

Documentation:
- Update the Okta changelog with a 2.10.4 release entry describing the system log deduplication fix.

Tests:
- Add a regression test to verify that both the checkpoint and events cache survive connector restarts using the shared data path.

Chores:
- Bump the Okta integration version from 2.10.3 to 2.10.4 in the manifest.

### DIFF
--- a/Okta/CHANGELOG.md
+++ b/Okta/CHANGELOG.md
@@ -7,11 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-06-16 - 2.10.4
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


mchupeau-sk (approved):
LGFM

PR #2802, opened 2026-06-17
fix(salesforce): change the run to async

TOUFIKIzakarya:
Main issue : [issue](https://github.com/SekoiaLab/platform/issues/89977)

## Summary by Sourcery

Make the Salesforce connector run loop fully asynchronous and delegate waiting to the connector instead of the timestepper.

New Features:
- Introduce an async_run entry point on the Salesforce connector and wrap it with the existing synchronous run method.

Bug Fixes:
- Ensure the Salesforce connector waits asynchronously when caught up instead of blocking the event loop within the timestepper ranges generator.
- Prevent the timestepper ranges generator from sleeping internally so it no longer blocks when a large sleep_duration is configured.
- Close the HTTP session and execute shutdown hooks when the connector stops running.

Tests:
- Add async_run tests for the Salesforce connector covering window iteration, non-blocking waits, and error handling for refresh token and generic exceptions.
- Add a timestepper test verifying that ranges() does not block even when sleep_duration is large.

### DIFF
--- a/Salesforce/salesforce/connector.py
+++ b/Salesforce/salesforce/connector.py
@@ -254,19 +254,20 @@ class SalesforceConnector(AsyncConnector):
 
         return result
 
-    def run(self) -> None:  # pragma: no cover
-        """Runs Salesforce connector with timestepper."""
+    async def async_run(self) -> None:  # pragma: no cover
+        """Runs Salesforce connector with timestepper asynchronously."""
         while self.running:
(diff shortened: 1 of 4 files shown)


mchupeau-sk (approved):
LGFM

mchupeau-sk:
@TOUFIKIzakarya just need to fix the CI

PR #2827, opened 2026-06-29
handle Graph PoolTimeout by resetting client and add regression test

tfanouillere-sekoia:
https://github.com/SekoiaLab/integration/issues/1707

## Summary by Sourcery

Handle Graph API client timeouts more robustly by resetting the client and add coverage for pool timeout handling.

Bug Fixes:
- Reset the Microsoft Graph client when timeout-related exceptions occur during async_run to avoid stuck or invalid HTTP transports.

Enhancements:
- Introduce a reusable _reset_graph_client helper to centralize Graph client shutdown and cleanup logic.
- Ensure the Graph API client closes its internal request adapter, including awaiting async close methods when present.

Tests:
- Add an async_run regression test verifying PoolTimeout is handled without logging an exception, logs a warning, and resets the Graph client.


mchupeau-sk:
@copilot update the version in the manifest.json file and also update the changelog with the changes on this PR

mchupeau-sk:
@tfanouillere-sekoia can you fix the pipeline errors ? (normally just mypy)

MicrosoftEntraID/tests/graph_api/test_client.py
```diff
@@ -101,3 +102,35 @@ async def test_client_get_directory_audits_empty_1(graph_api_client: GraphApi, d
...
+    credentials_close = AsyncMock()
+    graph_api_client._client.request_adapter.close = adapter_close
+    graph_api_client._credentials.close = credentials_close
+
+    await graph_api_client.close()
+
+    adapter_close.assert_called_once()
+    credentials_close.assert_awaited_once()
+    assert graph_api_client._client is None
+    assert graph_api_client._credentials is None
```
squioc, line 141: If you apply the requested changes, these tests will be useless; So, to be delete also.

squioc (changes requested):
Thank you for the work.

Some changes to apply

mchupeau-sk:
## 2. Blockers (Must Fix Before Merge)

### 2.1 `GraphApi.close()` doesn't actually close anything
**File**: [MicrosoftEntraID/graph_api/client.py](MicrosoftEntraID/graph_api/client.py#L131-L136)

The only change to this method in the PR is removing `# pragma: no cover`:
```python
async def close(self) -> None:
    if self._client:
        self._client = None
    if self._credentials:
        await self._credentials.close()
        self._credentials = None
```
Setting `self._client = None` merely drops the Python reference to the `GraphServiceClient`; it never closes its underlying `GraphRequestAdapter`/httpx transport. Relying on garbage collection to close an async httpx client is unsafe and can leave pooled connections open — which is exactly the condition that produces `httpx.PoolTimeout`. This is also the substance of an **unresolved** review comment from `squioc` on `tests/graph_api/test_client.py` ("If you apply the requested changes, these tests will be useless; so, to be deleted also."), which indicates the maintainer is still waiting on the real adapter-closing implementation described in the PR body.

**Proposed fix**: actually close/await-close the underlying transport inside `close()`, guarded so `_client`/`_credentials` are always cleared even if the underlying close raises. The exact attribute name for the kiota HTTP client varies by `msgraph-sdk`/`kiota-http` version, so probe for it defensively rather than hard-coding a private attribute:
```python
async def close(self) -> None:
    if self._client is not None:
        try:
            request_adapter = getattr(self._client, "request_adapter", None)
            http_client = getattr(request_adapter, "http_client", None)
            if http_client is not None and hasattr(http_client, "aclose"):
                await http_client.aclose()
        finally:
            self._client = None

    if self._credentials:
        try:
            await self._credentials.close()
        finally:
            self._credentials = None
```
Then add a test that exercises the real body of `close()` directly (not just via a monkey-patched `close` on the connector fixture), e.g. asserting that a fake `request_adapter.http_client.aclose` is awaited and that `_client`/`_credentials` end up `None` even when `aclose()` raises.

### 2.2 Unresolved `CHANGES_REQUESTED` review
The PR has an active, unaddressed `CHANGES_REQUESTED` review from `squioc` (2026-07-06, "Some changes to apply") and GitHub reports `mergeable_state: "blocked"`. Do not merge until this is resolved.

**Proposed fix**: no code change — reply to `squioc`'s review thread once 2.1 is implemented (and once `tests/graph_api/test_client.py` is updated/trimmed to match, as the reviewer requested), then re-request review.

---

## 3. Should-Fix (Convention/Correctness Issues)

### 3.1 `httpx` used directly but not declared as a dependency
**File**: [MicrosoftEntraID/pyproject.toml](MicrosoftEntraID/pyproject.toml#L33-L38)

`connector_entraid_graph_api.py` now does `import httpx` and relies on `httpx.TimeoutException`/`httpx.PoolTimeout` directly in production code, but `httpx` is not listed under `[tool.poetry.dependencies]` — it currently resolves only transitively via `msgraph-sdk`/`azure-identity`. Add it as an explicit dependency so a future transitive-dependency change can't silently break this import.

**Proposed fix**:
```diff
 [tool.poetry.dependencies]
 python = ">=3.11,<3.12"
 azure-identity = "^1.23.1"
 pydantic = "2.10"
 msgraph-sdk = "^1.45.0"
 cachetools = "^6.2.0"
 loguru = "^0.7.3"
+httpx = "^0.27"
```
(pin the actual resolved transitive version from `poetry.lock` to avoid an unintended upgrade), then run `poetry lock` to refresh `poetry.lock`.

### 3.2 CHANGELOG entry left under `## Unreleased` despite a version bump in the same PR
**File**: [MicrosoftEntraID/CHANGELOG.md](MicrosoftEntraID/CHANGELOG.md#L9-L13)

The new "Fixed" bullets are added under `## Unreleased`, but [manifest.json](MicrosoftEntraID/manifest.json#L44) is bumped to `2.10.24` in the same PR. Every other module in the repo (e.g. `Sophos/CHANGELOG.md`) keeps `## Unreleased` empty and instead adds a dated `## <date> - <version>` section matching the manifest bump. Move these bullets into a new `## 2026-07-07 - 2.10.24` section for consistency with the rest of the repo.

**Proposed fix**:
```diff
 ## Unreleased
 
-### Fixed
-
-- Reset the Microsoft Graph client when timeout-related exceptions occur in the asset connector loop
-- Add a regression test for Graph `PoolTimeout` handling and client reset behavior
-
+## 2026-07-08 - 2.10.24
+
+### Fixed
+
+- Reset the Microsoft Graph client when timeout-related exceptions occur in the asset connector loop
+- Add a regression test for Graph `PoolTimeout` handling and client reset behavior
+
 ## 2026-06-16 - 2.10.23
```

### 3.3 Regression test only checks that `log` was called, not what was logged
**File**: [MicrosoftEntraID/tests/asset_connector/test_microsoft_entraid_graph_api.py](MicrosoftEntraID/tests/asset_connector/test_microsoft_entraid_graph_api.py#L83)

`test_entraid_connector_async_run_handles_pool_timeout` and the two similar tests call `entraid_connector.log.assert_called()`, which would pass even for an unrelated log call. This mirrors an unresolved review comment from `copilot-pull-request-reviewer`. Tighten to assert the actual message/level for the PoolTimeout warning.

**Proposed fix**:
```python
    await entraid_connector.async_run()

    entraid_connector.log_exception.assert_not_called()
    entraid_connector.log.assert_called_once()
    _, kwargs = entraid_connector.log.call_args
    assert kwargs["level"] == "warning"
    assert "PoolTimeout" in kwargs["message"]
    mocked_close.assert_awaited_once()
    assert entraid_connector._client is None
```
apply the equivalent tightened assertion to `test_entraid_connector_async_run_handles_timeout_exception` and `test_entraid_connector_async_run_handles_asyncio_timeout` (checking `TimeoutException`/`TimeoutError` in the message respectively).

---

## 4. Nits (Minor/Optional Improvements)

### 4.1 Exception Hierarchy Documentation
**File**: [MicrosoftEntraID/azure_ad/connector_entraid_graph_api.py](MicrosoftEntraID/azure_ad/connector_entraid_graph_api.py#L206)

**Current code**:
```python
except (TimeoutError, asyncio.TimeoutError, httpx.TimeoutException) as error:
```

**Observation**: The PR description and test names specifically emphasize `httpx.PoolTimeout`, but the production code catches the base class `httpx.TimeoutException`. While this is technically correct (PoolTimeout inherits from TimeoutException and is properly caught), consider adding a clarifying comment:

```python
# Catches TimeoutError, asyncio.TimeoutError, and all httpx timeout subclasses
# (including PoolTimeout, ConnectTimeout, ReadTimeout, WriteTimeout)
except (TimeoutError, asyncio.TimeoutError, httpx.TimeoutException) as error:
```

**Justification**: The test coverage confirms this works correctly. Catching the base class is better Python practice than enumerating all subclasses. A comment would help future maintainers understand the inheritance relationship without needing to consult httpx documentation.

**Proposed fix**: extract the tuple into a module-level constant so intent is documented in one place and reused if other methods need the same check:
```python
GRAPH_TIMEOUT_EXCEPTIONS = (TimeoutError, httpx.TimeoutException)  # asyncio.TimeoutError is TimeoutError on py311+
...
except GRAPH_TIMEOUT_EXCEPTIONS as error:
    ...
```

### 4.2 Wrap the reset call defensively in `async_run`
As noted by Sourcery's automated review, once `close()` performs real I/O (per blocker 2.1) it has more ways to raise; consider a narrow `try/except` around `await self._reset_graph_client()` in the timeout handler so a failure during cleanup can't escape and break the main `async_run` loop.

**Proposed fix**:
```python
except GRAPH_TIMEOUT_EXCEPTIONS as error:
    self.log(
        message=f"A timeout was raised by the client ({type(error).__name__}), resetting Graph client",
        level="warning",
    )
    try:
        await self._reset_graph_client()
    except Exception as reset_error:
        self.log(message=f"Failed to reset Graph client cleanly: {reset_error}", level="warning")
    await asyncio.sleep(self.configuration.frequency)
```

squioc (approved):
LGTM

PR #2855, opened 2026-07-03
prevent CloudTrail connector from staying in starting

tfanouillere-sekoia:
https://github.com/SekoiaLab/platform/issues/89004

## Summary by Sourcery

Update AWS abstract connector run loop to maintain liveness and handle batches without source timestamps.

Enhancements:
- Refresh connector heartbeat on each processing cycle so supervisor liveness reflects an active loop.
- Handle successful batches with missing message timestamps by recording a default metric instead of computing lag.

Tests:
- Add a regression test ensuring the run loop handles empty timestamp lists without errors and updates the heartbeat.


squioc:
Hello

The proposal won't fix the root cause of the issue as the issue was on a lower level than the connector level.
I close the PR consequently.

Regards.

PR #2865, opened 2026-07-07
feat(harfanglab): add feature flag

TOUFIKIzakarya:
Linked issue : [issue](https://github.com/SekoiaLab/platform/issues/88780#issuecomment-4903791235)

Test with the first asset connector

## Summary by Sourcery

Add a feature-flag capability to the HarfangLab software asset connector and bump its version.

New Features:
- Introduce a feature-flag configuration on the HarfangLab software asset connector.

Documentation:
- Document the new feature-flag support in the HarfangLab changelog.

### DIFF
--- a/HarfangLab/CHANGELOG.md
+++ b/HarfangLab/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-07-07 - 1.31.7
+
+### Changed
+
+- Add `feature-flag` to the software asset connector
+
 ## 2026-07-03 - 1.31.6
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


StephKll3c:
Thanks for the quick turnaround. One suggestion before this lands: `feature-flag` as a boolean won't scale once more than one gated rollout exists at the same time. #88780 is specifically about scoping the software asset connector to EAP communities via the `reveal-application-discovery` flag, and the backend design in SekoiaLab/platform#89206 has `GET /v1/symphony/connectors` receiving the caller's active PostHog flags and filtering connectors by name, not by a plain true/false.

There's also the separate `beta-flag` idea from the discussion on #88780, for letting clients try a new integration of an existing connector type in production ahead of GA. Both cases can be represented the same way instead of as two separate fields.

Could we make this a list instead, something like `"feature-flags": ["reveal-application-discovery"]`, sitting next to `capability` the way it already does here. That way a connector can declare zero, one, or several flags, including something like `reveal-beta-connectors` for the beta case, and the backend can just check for overlap with whatever flags the frontend sends. Happy to help wire this up if useful.

TOUFIKIzakarya:
Hello @StephKll3c,

Thanks for the comment, it explain clearly what we want. I have done the modification.

StephKll3c (approved):

PR #2866, opened 2026-07-07
Sekoia.io: Guard GetAsset.run() against empty asset_uuid

PierrickV:
## Summary
- `GetAsset.perform_request()` was called even when `asset_uuid` was empty/missing, issuing an HTTP GET with an empty uuid segment in the URL.
- Added an early guard in `GetAsset.run()`: if `asset_uuid` is empty, call `self.error("Asset uuid is required")` and return `None` immediately, without invoking `perform_request()`.
- Added a unit test asserting `run()` returns `None` and no HTTP request is made when `uuid` is empty.

## Testing
- `poetry run pytest tests/operation_center_action/test_get_asset.py` — all 4 tests pass.

## Summary by Sourcery

Guard the Sekoia.io GetAsset action against empty asset UUID inputs to avoid issuing invalid HTTP requests and surface a clear error instead.

Bug Fixes:
- Prevent GetAsset from performing an HTTP request when the provided asset UUID is empty or missing.

Tests:
- Add a unit test verifying GetAsset.run() returns None and performs no HTTP requests when called with an empty UUID argument.

### DIFF
--- a/Sekoia.io/CHANGELOG.md
+++ b/Sekoia.io/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-07-07 - 2.74.1
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


mchupeau-sk (approved):
LGFM

PR #2873, opened 2026-07-09
Sekoia.io: Validate GetAlert arguments via Pydantic

PierrickV:
## What

Validate `GetAlert`'s `uuid` argument via a Pydantic v2 model before the action runs, so a missing or malformed alert UUID fails immediately with a clear validation error instead of the action issuing an HTTP request with bad/empty data.

## Changes

- `uuid` → `uuid.UUID` (rejects missing, blank, and invalid-shape values natively)

## Testing

- `pytest`, `black`, and `mypy` all pass.


Sekoia.io/sekoiaio/operation_center/get_alert.py
```diff
@@ -0,0 +1,29 @@
+from pydantic.v1 import BaseModel, Field, validator
```
squioc: Can you update the sekoia-automation-sdk to 1.23.0 that supports Pydantic V2 completely? (see [changelog](https://github.com/SEKOIA-IO/sekoia-automation-sdk/blob/main/CHANGELOG.md#1230---2026-05-12))

squioc (changes requested):
Thank you for the work.

Can you upgrade the sekoia-automation-sdk dependency to the last version in order to fully se Pydantic V2?
Thank you

PierrickV:
Thanks for the review — addressed:

- Upgraded `sekoia-automation-sdk` to `1.23.1` and migrated `GetAlert` from the `pydantic.v1` shim to native Pydantic v2 (`field_validator`, `model_dump`).
- This SDK bump surfaced a latent "Mixing V1 models and V2 models" error (previously just a warning) in 4 other actions in this module still on the v1 shim (`CreateDataset`, `DeleteDataset`, `ExecuteAQuery`, `ListQueries`) — migrated those to native Pydantic v2 too so the whole test suite passes cleanly again (315 passed, 0 warnings).
- Also applied Sourcery's suggestions: centralized the duplicated `base_url` constant into `sekoiaio/operation_center/constants.py`, and added the `stix` query parameter to the happy-path test.

All tests, mypy, and black checks pass.

Sekoia.io/sekoiaio/operation_center/get_alert.py
PierrickV: Applied @squioc

PierrickV:
Updated this PR to replace the custom `uuid` "must not be blank" validator with a native Pydantic `uuid.UUID` field. This gives stronger validation because it now rejects any non-UUID-shaped string, not just blank values. I also added a dedicated invalid-shape test case (`not-a-uuid`) to ensure no HTTP request is sent when validation fails.

mchupeau-sk:
Hello @PierrickV and @squioc, I don't agree with this solution, for me there is more simply way to do the guard.

We can just update the get alert json config file and adding the following parameter  :

```
        "pattern": {
            "type": "string",
            "format": "regex for uuid v4"
        }
```

PierrickV:
@mchupeau-sk:
- is the json config validated at runtime or only when you configure your action ?
- on maintaining a regex here vs. the Pydantic model: Pydantic is our standard for argument validation in Python actions
 across this codebase, so I'd rather not introduce/maintain hand-written regexes as a parallel validation mechanism when Pydantic
 already gives us built-in, well-tested types (`UUID`, etc.) with clear error messages and is the pattern reviewers/contributors 
already expect here. It also keeps validation logic in one place (the code) instead of split between the manifest and the action.

mchupeau-sk (approved):
LGFM

mchupeau-sk:
need @squioc approvment and we can go for qualification

squioc (approved):
LGTM

PR #2874, opened 2026-07-09
Whois: Validate query argument via Pydantic

PierrickV:
## What

Validate Whois's `query` argument via a Pydantic v2 model before the action runs, so a missing or blank query fails immediately with a clear validation error instead of performing a WHOIS lookup with bad data.

## Changes

- `query` → `NonEmptyStr` (kept as a plain non-empty string since it accepts a domain, URL, or IP address interchangeably)

## Testing

- `pytest` and `black` pass.


PierrickV:
SDK bumped `1.13.0` → `1.23.1`; migrated `Whois/whois_module/whois_action.py` and `Whois/tests/test_whois_action.py` from the `pydantic.v1` shim to native Pydantic v2.

Validation summary: `pytest tests -q` passed 7/7 tests before push and 7/7 tests after the final rebase; `poetry run black --check whois_module tests` also passes locally. Mypy is not configured for the Whois module, so there was no mypy run to report.

PierrickV:
Simplified the WhoisArguments.query validation by replacing the custom blank-check field validator with a NonEmptyStr alias backed by Pydantic v2 StringConstraints(strip_whitespace=True, min_length=1).

Importantly, query was deliberately kept as a plain non-empty string rather than tightened to a more specific type such as an IP or URL field, because this argument is intentionally polymorphic and accepts domain names, URLs, or IP addresses interchangeably.

PierrickV:
Consolidated the CHANGELOG: an earlier phase had accidentally stacked two separate version entries (guard fix + SDK bump) for this still-unreleased PR. Merged them into a single `1.25.1` entry combining both the `Fixed` and `Changed` notes.

PR #2875, opened 2026-07-09
Mattermost: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate Mattermost's action arguments via Pydantic v2 models before each action runs, so missing/blank/malformed input fails immediately with a clear validation error instead of the action posting to a bad URL or with empty data.

## Changes

- `alert_uuid` → `uuid.UUID`
- `base_url` → `HttpUrl`
- `api_key`, `message` → `NonEmptyStr`

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to the native Pydantic v2 pattern used by the other SDK-upgrade branches. SDK bumped 1.13.0→1.23.1. Migrated files: Mattermost/mattermost/action_mattermost_postalert.py, Mattermost/mattermost/action_mattermost_postmessage.py, Mattermost/tests/test_action_mattermost_postalert.py, Mattermost/tests/test_action_mattermost_postmessage.py. Test suite size stayed 9→9, with 9/9 passing locally. black passes, and mypy passes locally with --ignore-missing-imports (the module has no dedicated mypy config and still depends on untyped third-party packages).

PierrickV:
Updated Mattermost entrypoint typing to rely on native Pydantic constraints instead of shared blank-check validators. Upgraded fields: postalert.alert_uuid -> UUID (validated RFC-4122 identifier), postalert.base_url -> HttpUrl (validated HTTP(S) API base URL). Kept as non-empty strings: postalert.api_key -> NonEmptyStr (opaque secret token, not a structured URL/UUID), postmessage.message -> NonEmptyStr (free-form user text). Also adjusted URL building to cast typed values back to str where needed and added invalid-shape tests for alert_uuid/base_url.

PR #2876, opened 2026-07-09
Sekoia.io: Validate UpdateAsset/UpdateAlertStatus arguments via Pydantic

PierrickV:
## What

Validate Sekoia.io's `UpdateAsset` and `UpdateAlertStatus` action arguments via Pydantic v2 models before each action runs, so a missing/malformed UUID or blank status fails immediately with a clear validation error instead of issuing an HTTP request with bad data.

## Changes

- `uuid` → `uuid.UUID` (both actions)
- `status` → `NonEmptyStr` (kept as a free-form string since customers can now define custom alert statuses, not just the built-in set)

## Testing

- `pytest`, `black`, and `mypy` all pass.


PierrickV:
Mirroring the Pydantic v2 follow-up from sibling PR #2873: bumped `sekoia-automation-sdk` from `^1.21.1` to `^1.23.1`, regenerated the lockfile, and migrated the remaining operation-center `pydantic.v1` usages in `update_asset.py`, `update_alert_status.py`, `assets_merge.py`, `synchronize_assets_with_ad.py`, `create_dataset.py`, `delete_dataset.py`, `execute_a_query.py`, and `list_queries.py` (plus the related tests) to native Pydantic v2 imports/validators.

Validation on this branch is green after the bump: `poetry run pytest tests -q` (314 passed, 1 skipped), `poetry run mypy sekoiaio --ignore-missing-imports` (clean; only existing informational notes), and `poetry run black --check sekoiaio tests` all pass, and there are no remaining `Mixing V1 models and V2 models` warnings in the test run.

PierrickV:
Updated both `UpdateAsset` and `UpdateAlertStatus` to validate their `uuid` arguments as native Pydantic `UUID` values, so empty and malformed UUID strings are rejected before any HTTP call. I also added explicit invalid-shape UUID test coverage for both actions.

For `UpdateAlertStatus.status`, I deliberately kept it as a free-form non-empty string (using `StringConstraints` only) rather than a fixed enum/Literal, because the platform supports custom/user-defined alert statuses beyond the historical built-in set.

PR #2877, opened 2026-07-09
Ilert: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate Ilert's `TriggerAlert` action arguments via a Pydantic v2 model before the action runs, so missing/blank/malformed input fails immediately with a clear validation error instead of the action posting to a bad URL or with empty data.

## Changes

- `alert_uuid` → `uuid.UUID`
- `base_url` → `HttpUrl`
- `api_key` → `NonEmptyStr`

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to the same native Pydantic v2 pattern used in the Sekoia.io module:\n\n- bumped sekoia-automation-sdk from the old ^1.13.0 pin to ^1.23.1\n- migrated pydantic.v1 usages in ilert/action_ilert_trigger_alert.py and tests/test_action_ilert_triggeralert.py to native Pydantic v2\n- refreshed poetry.lock and aligned the module Python constraint with the SDK's Python >=3.11 requirement\n\nValidation: poetry run pytest tests -q, poetry run mypy ilert --ignore-missing-imports, and poetry run black --check ilert tests all pass; no pydantic v1/v2 model-mixing warnings remain.

PierrickV:
Updated the trigger-alert argument model to use stronger Pydantic-native types: alert_uuid is now parsed as UUID, base_url as HttpUrl, and api_key remains a stripped non-empty string via StringConstraints. This removes the shared field validator, lets Pydantic enforce UUID/URL shape directly, preserves the existing empty/blank rejection behavior, and adds regression coverage for invalid UUID and URL inputs.

PR #2878, opened 2026-07-09
IKnowWhatYouDownload: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate IKnowWhatYouDownload's action arguments via Pydantic v2 models before each action runs, so a missing/blank/malformed IP or CIDR fails immediately with a clear validation error instead of issuing an API call with bad data.

## Changes

- `ip` → `IPvAnyAddress` (ipexist, iphistory)
- `cidr` → `IPvAnyNetwork` (iplist)

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to finish the native Pydantic v2 migration:

- bumped `sekoia-automation-sdk` to `1.23.1` (latest; this is the first line where the SDK wraps actions with native `pydantic.validate_call`, so it removes the v1/v2 model-mixing failure)
- migrated `iknowwhatyoudownload/action_iknow_ipexist.py`, `iknowwhatyoudownload/action_iknow_iphistory.py`, `iknowwhatyoudownload/action_iknow_iplist.py` and their tests off the `pydantic.v1` shim
- added `tests/conftest.py` so `poetry run pytest tests -q` imports the module correctly in this package-mode=false layout
- validation checks: `pytest tests -q`, `mypy iknowwhatyoudownload --ignore-missing-imports`, and `black --check iknowwhatyoudownload tests` all pass

PierrickV:
Updated the IKnowWhatYouDownload entrypoint validation to use native Pydantic network types:
- IKnowIPExistArguments.ip -> IPvAnyAddress
- IKnowIPHistoryArguments.ip -> IPvAnyAddress
- IKnowIPListArguments.cidr -> IPvAnyNetwork

Also added invalid-shape validation coverage for the action tests (for example non-IP values for ip, and non-CIDR values for cidr) while keeping the existing blank/empty ValidationError assertions.

PierrickV:
Consolidated the CHANGELOG: merged the two stacked version entries (`1.24.1`/`1.24.2`) into a single `1.24.1` entry.

PR #2879, opened 2026-07-09
CrowdStrikeFalcon: Validate entrypoint argument via Pydantic

PierrickV:
## What

Validate CrowdStrikeFalcon's IOC-upload action argument via a Pydantic v2 model before the action runs, so a missing/blank IOC value fails immediately with a clear validation error instead of uploading an empty indicator.

## Changes

- `value` → `NonEmptyStr` (kept as a plain non-empty string since its valid shape is polymorphic — IP, domain, or hash — depending on the sibling `type` field)

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to bump sekoia-automation-sdk from 1.22.5 to 1.23.1 and finish the native Pydantic v2 migration needed with the newer validate_call behavior. Migrated the remaining shim usages in crowdstrike_falcon/custom_iocs.py, crowdstrike_falcon/models.py, and tests/test_actions_iocs.py, and also switched the local Pydantic v2 deprecations in crowdstrike_falcon/client/schemas.py and crowdstrike_falcon/client/__init__.py to ConfigDict/model_dump. Validation is green: poetry run pytest tests -q (154 passed, 5 skipped), poetry run mypy crowdstrike_falcon --ignore-missing-imports, and poetry run black --check crowdstrike_falcon tests.

PierrickV:
Simplified the Pydantic guard in `CrowdstrikeActionAddIOCArguments` by replacing the custom `@field_validator` blank check with a shared `NonEmptyStr = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]`.

I deliberately kept `value` as a plain non-empty string instead of tightening it to an IP/hash/domain type because its valid shape is polymorphic and depends on the sibling IOC `type` (for example `ipv4`, `domain`, `sha256`, etc.). Modeling that safely would require additional cross-field/discriminated validation logic, which felt out of scope for this cleanup.

I also left `type` as a plain string: the accepted values differ by action and are already checked against each action's supported type list at runtime, so introducing a stricter shared enum/Literal here would be a broader behavior change rather than a validator simplification.

PierrickV:
Consolidated the CHANGELOG: merged the two stacked version entries (`1.25.18`/`1.25.19`) from an earlier phase into a single `1.25.18` entry.

PR #2880, opened 2026-07-09
TheHiveV5: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate TheHiveV5's action arguments via Pydantic v2 models before each action runs, so missing/blank required input fails immediately with a clear validation error instead of the action calling the API with bad data.

## Changes

- `alert_id`, `message`, `filepath` → `NonEmptyStr` (kept as plain non-empty strings — TheHive alert IDs use an opaque `~`-prefixed format, not a UUID)

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to migrate TheHiveV5 to native Pydantic v2 validation and bump the SDK from 1.13.0 to 1.23.1.

Migrated files:
- TheHiveV5/thehive/add_commment.py
- TheHiveV5/thehive/add_observable.py
- TheHiveV5/thehive/upload_logs.py

Local validation:
- pytest remains 43/43 passing (count unchanged before/after)
- black passes
- this module has no dedicated mypy configuration; an ad hoc mypy run only reports pre-existing missing-stub import issues

CI note: codecov/project still fails overall, but codecov/patch/TheHive and codecov/project/TheHive are both passing, so the remaining repo-wide coverage failure looks unrelated to this module.

PierrickV:
Simplified the TheHiveV5 argument guards by replacing the custom blank-check @field_validator methods with a shared Pydantic v2 NonEmptyStr alias (Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]). I intentionally kept these fields as non-empty strings rather than tightening their types: TheHive alert_id uses TheHive's own ~-prefixed internal identifier format (for example ~12345, not an RFC-4122 UUID), add_comment.message is free text, and upload_logs.filepath is a local path string. In add_observable, the only required opaque string field here is alert_id; the observable payload itself stays polymorphic via the existing events structure.

PR #2881, opened 2026-07-09
Sophos: Validate endpoint_id via Pydantic across EDR actions

PierrickV:
## What

Validate Sophos EDR action arguments via a shared Pydantic v2 model before each action runs, so a missing/malformed endpoint ID fails immediately with a clear validation error instead of calling the API with bad data.

## Changes

- `endpoint_id` → `uuid.UUID` (isolate, deisolate, run scan)
- Removed an empty `SophosEndpointScanArguments` subclass that added no fields beyond its parent

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to native Pydantic v2 validation and bumped sekoia-automation-sdk from 1.22.5 to 1.23.1. Migrated: Sophos/sophos_module/action_sophos_edr_isolate.py, Sophos/sophos_module/action_sophos_edr_deisolate.py, Sophos/sophos_module/action_sophos_edr_run_scan.py, plus the Sophos configuration model in Sophos/sophos_module/base.py. Local validation stayed at 207 passed / 3 skipped on the verification pytest runs, and black, mypy, and pytest all pass.

PierrickV:
Updated the three Sophos EDR endpoint actions to validate endpoint_id as a native Pydantic UUID and to stringify it at API call sites. I also removed the old blank-string validators, amended the changelog entry, and extended test coverage with an invalid UUID shape case so the suite now checks both missing or blank and malformed endpoint IDs.

PierrickV:
Removed the empty `SophosEndpointScanArguments(SophosEndpointArguments): pass` subclass — it added no new fields/validators beyond its parent, so `action_sophos_edr_run_scan.py` now types its `run()` argument as `SophosEndpointArguments` directly. Tests (208 passed, 3 skipped) and `black` still clean.

Sophos/sophos_module/base.py
```diff
@@ -10,7 +12,11 @@ class SophosConfiguration(BaseModel):
         description="API Url of the Sophos Central API (e.g. 'https://api-{dataRegion}.central.sophos.com')",
     )
     client_id: str = Field(..., description="OAuth2 client identifier")
-    client_secret: str = Field(secret=True, description="OAuth2 client secret")
+    client_secret: str = Field(..., description="OAuth2 client secret", json_schema_extra={"secret": True})
+
+
+class SophosEndpointArguments(BaseModel):
+    endpoint_id: UUID = Field(..., description="Endpoint ID")
```
mchupeau-sk: @PierrickV endpointID is not UUID type

PR #2882, opened 2026-07-09
HarfangLab: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate HarfangLab's action arguments via Pydantic v2 models before each action runs, so a missing/malformed agent ID or IP fails immediately with a clear validation error instead of calling the API with bad data.

## Changes

- `agent_id` → `uuid.UUID` (get_agent_telemetry)
- `target_ip` → `IPvAnyAddress` (get_hostnames_by_ip)

## Testing

- `pytest` and `ruff format` pass.

### DIFF
--- a/HarfangLab/CHANGELOG.md
+++ b/HarfangLab/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-07-09 - 1.31.7
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


PierrickV:
Updated the HarfangLab entrypoint models to use stronger Pydantic-native types:
- agent_id is now a UUID, so malformed identifiers (for example not-a-uuid) fail validation before we build telemetry requests.
- target_ip is now an IPvAnyAddress, so malformed addresses (for example not-an-ip) fail validation before we query the Agent endpoint.
- Request serialization now wraps both typed values with str(...) at the HTTP boundary.

I also added regression tests for invalid UUID/IP shapes. I considered converting alert_created to a datetime field too, but left it as a string for now to keep this PR scoped to the two confirmed argument semantics.

PR #2883, opened 2026-07-09
MicrosoftOutlook: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate MicrosoftOutlook's action arguments via Pydantic v2 models before each action runs, so missing/blank required input fails immediately with a clear validation error instead of calling the Graph API with bad data.

## Changes

- `user`, `message_id` → `NonEmptyStr` (kept as plain non-empty strings — `user` accepts either an opaque object ID or a UPN, and `message_id` is an opaque Graph identifier)

## Testing

- `pytest` and `black` pass.


PierrickV:
Bumped `sekoia-automation-sdk` to `1.23.1` and migrated all MicrosoftOutlook native Pydantic usage away from the `pydantic.v1` shim. Updated files: `microsoft_outlook_modules/action_get_message.py`, `action_delete_message.py`, `action_forward_message.py`, `action_update_message.py`, `action_send_message.py`, `models.py`, plus `tests/test_actions.py`, `pyproject.toml`, `poetry.lock`, and `CHANGELOG.md`. Validation status: `poetry run pytest tests -q` ✅ (36 passed), `poetry run mypy microsoft_outlook_modules --ignore-missing-imports` ✅, `poetry run black --check microsoft_outlook_modules tests` ✅. The send/update `from`/`from_` alias behavior is now explicitly covered in tests and still passes.

PierrickV:
Simplified the message action argument models by replacing repeated blank-check field_validator methods with a shared NonEmptyStr Pydantic constraint in microsoft_outlook_modules.models. This keeps the same missing/blank validation behavior with less custom code.

I intentionally kept user as a non-empty string because Microsoft Graph accepts either an opaque user object ID or a UPN/email-like value, so EmailStr would wrongly reject valid object IDs. I also kept message_id as a non-empty string because Graph message IDs are opaque identifiers, not UUIDs.

I left recipient-like fields as plain strings as well: email-validator is not declared in MicrosoftOutlook/pyproject.toml, so upgrading them to EmailStr here would add a new dependency.

PR #2884, opened 2026-07-09
Virustotal: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate Virustotal's action arguments via Pydantic v2 models before each action runs, so missing/blank/malformed input fails immediately with a clear validation error instead of calling the API with bad data.

## Changes

- `ip` → `IPvAnyAddress` (scanip)
- `url` → `HttpUrl` (scanurl)
- `resource`, `comment`, `file`, `hash` → `NonEmptyStr` (kept as plain non-empty strings — no single stricter type fits their polymorphic/opaque shape)

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR for the Pydantic v2 / SDK migration.\n\n- SDK bump: sekoia-automation-sdk 1.13.0 → 1.23.1\n- Migrated to native Pydantic v2 field_validator: Virustotal/virustotal/action_virustotal_getcomments.py, Virustotal/virustotal/action_virustotal_postcomment.py, Virustotal/virustotal/action_virustotal_scanfile.py, Virustotal/virustotal/action_virustotal_scanhash.py, Virustotal/virustotal/action_virustotal_scanip.py, Virustotal/virustotal/action_virustotal_scanurl.py\n- Test count: 26 before / 26 after\n- Local checks: black ✅, mypy ✅, pytest ✅

PierrickV:
Updated the VirusTotal argument validation as discussed:

- Upgraded typed fields:
  - `ip` -> `IPvAnyAddress` in `action_virustotal_scanip.py`
  - `url` -> `HttpUrl` in `action_virustotal_scanurl.py`
- Kept as non-empty strings (using a shared `NonEmptyStr` alias with whitespace stripping):
  - `resource` in get/post comment actions, because VirusTotal accepts polymorphic resource identifiers (hash, URL, domain, etc.)
  - `comment`, because it is free text
  - `file`, because it is a local file path string in this codebase
  - `hash`, because there is no built-in Pydantic hash type and current fixtures do not require a stricter custom regex

Also removed the old blank-check `@field_validator` methods, updated the `ip`/`url` call sites to use `str(...)`, added invalid-shape tests for `ip="not-an-ip"` and `url="not a url"`, and re-ran the Virustotal test suite successfully (28 passed).

Virustotal/virustotal/types.py
```diff
@@ -0,0 +1,5 @@
+from typing import Annotated
+
+from pydantic import StringConstraints
+
+NonEmptyStr = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
```
mchupeau-sk, line 5: @PierrickV is not better to put it in the SDK to avoid to recreate it in each module ?

PR #2886, opened 2026-07-09
Onyphe: Validate entrypoint argument via Pydantic

PierrickV:
## What

Validate Onyphe's `datascan` action argument via a Pydantic v2 model before the action runs, so a missing/blank/malformed IP or search string fails immediately with a clear validation error instead of calling the API with bad data.

## Changes

- `ip` → `IPvAnyAddress | None`
- `string` kept as `NonEmptyStr | None`
- The existing cross-field validator (exactly one of `ip`/`string` must be set) is preserved unchanged

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to the native Pydantic v2 pattern used in the related rollout.

- SDK: sekoia-automation-sdk 1.13.0 -> 1.23.1
- Migrated file(s): Onyphe/onyphe/action_onyphe_datascan.py
- Tests: 78 -> 79 (added the whitespace-only invalid entrypoint case)
- Local checks: black ✅, pytest ✅ (79 passed), mypy run completed with only the module's pre-existing missing-stub/untyped-import issues remaining; no new migration-specific type error remains

PierrickV:
Updated the datascan argument model while intentionally leaving the existing cross-field XOR validator untouched. The optional ip field is now validated as IPvAnyAddress | None, with whitespace trimming and rejection of non-string inputs preserved. The string field now uses a non-empty constrained string, and the action stringifies the validated IP when building the request URL. Added coverage for an invalid-shape IP and the IP-only success path.

mchupeau-sk:
Hello @PierrickV, for me this PR is incorrect :/ 

**File:** `Onyphe/onyphe/action_onyphe_datascan.py`
**Lines:** 61–64 (the `run()` signature and `resource = str(arguments.ip) ...`)

````
This won't work at runtime — `run(self, arguments: OnypheDatascanArguments)` is only a type hint, so `arguments` is still the raw `dict` the SDK hands over. `arguments.ip` on line 63 raises `AttributeError`, not `ValidationError`.

The new tests confirm it: `test_ip_argument_uses_validated_ip_in_request_url` calls `action.run({"ip": "93.184.216.34"})` with a plain dict, and `generic_onyphe_tests.py` does `arguments.update({"budget": 2})` before calling `run()`. Nothing ever constructs the model.

Suggest following the pattern already used in `MicrosoftOutlook/microsoft_outlook_modules/action_get_message.py` and `HTTP/http_module/action_download_file.py` — keep the dict signature and validate explicitly:

```python
def run(self, arguments: Any) -> dict:
    validated = OnypheDatascanArguments.model_validate(arguments)

    resource = str(validated.ip) if validated.ip is not None else validated.string
    get_url = urljoin("https://www.onyphe.io/api/v2/simple/", "datascan/" + resource)

    return get_with_paging(
        get_url, self.module.configuration, validated.budget, {"page": validated.first_page}
    )
```

While you're in there, two smaller things:

`prepare_ip` can go away. I assume it exists because pydantic v2's lax mode accepts `int` for `IPvAnyAddress` (`8` → `0.0.0.8`), which would break the `{"ip": 8}` case. `model_config = ConfigDict(extra="ignore", strict=True)` gets you that rejection declaratively and applies it to every field, so `{"budget": "2"}` stops silently coercing too. The whitespace-stripping half I'd just drop — silently accepting `" 8.8.8.8 "` is arguably worse than rejecting it.

And `assert resource is not None` should go — the `model_validator` already guarantees it, and `assert` is stripped under `python -O`.

Also worth adding bounds while the model is being written: `budget: int = Field(default=1, ge=0)` and `first_page: int = Field(default=1, ge=1)`. Today a negative `first_page` is accepted and sent straight to Onyphe.
````

**Why this matters:** the annotation-vs-`arguments_model` distinction is easy to miss in review because the code *reads* as if validation happens. But the SDK passes whatever the playbook sent — a dict — and the first attribute access blows up. This means the action fails on *every* invocation, including the happy path, and the new tests would fail in CI rather than pass. It's worth checking whether CI actually ran on this PR, because if those tests are green, something else is going on that's worth understanding before merging.

Three more points you may want to raise separately, since they don't attach to this line:

- **`get_arg_ip` is still used by 8 sibling actions** (`ip`, `synscan`, `forward`, `reverse`, `inetnum`, `threatlist`, `sniffer`, `pastries`). The commit message says "validate action arguments via Pydantic" but only `datascan` moved, leaving two validation styles in one module.
- **The manifest is out of sync** — `action_onyphe_datascan.json` still has `"required": []` with no exclusivity constraint, contradicting the new model.

PR #2887, opened 2026-07-09
Fortigate: Validate entrypoint arguments via Pydantic

PierrickV:
## What

Validate Fortigate's action arguments via Pydantic v2 models before each action runs, so missing/blank/malformed input fails immediately with a clear validation error instead of calling the API with bad data.

## Changes

- `ip` → `IPvAnyAddress` (add_ip_address)
- `fqdn`, `name` → `NonEmptyStr` (kept as plain non-empty strings — no built-in stricter Pydantic type exists for FQDNs/object names)

## Testing

- `pytest` and `black` pass.


PierrickV:
Updated this PR to use native Pydantic v2 end-to-end by upgrading sekoia-automation-sdk to 1.23.1 and aligning the module's Python requirement with the SDK (>=3.11,<3.12).

Migrated files:
- fortigate/action_fortigate_add_ip_address.py
- fortigate/action_fortigate_add_group_address.py
- fortigate/action_fortigate_disable_local_user.py
- fortigate/action_fortigate_add_fqdn.py
- tests/test_action_fortigate_addaddress.py
- tests/test_action_fortigate_addgroup.py
- tests/test_action_fortigate_disablelocaluser.py
- tests/test_action_fortigate_addfqdn.py

Validation: pytest (17 passed), mypy passed, black --check passed.

PierrickV:
Updated the Fortigate argument models so add_ip_address.ip now uses Pydantics IPvAnyAddress, which preserves the existing non-empty guard and also rejects invalid IP shapes like "not-an-ip" before we build the Fortigate payload. The FQDN and opaque Fortigate object/user names were kept as trimmed non-empty strings because Pydantic core does not provide a built-in FQDN type and those name fields are Fortigate-specific identifiers rather than standard network primitives.

mchupeau-sk:
@PierrickV if you can resolve the conflicts

PR #2906, opened 2026-07-20
fix flare

mchupeau-sk:
## Summary by Sourcery

Streamline Flare events connector paging to comply with Flare v4 API limits while improving throughput and simplifying retry behavior.

Bug Fixes:
- Align event pagination with the Flare v4 events search endpoint and enforce the maximum allowed page size.

Enhancements:
- Stream events to the intake in batches of 100 while advancing the checkpoint only on page boundaries to improve performance and resilience.
- Use the configured throttle setting for both page and event requests and remove the custom HTTP session in favor of the SDK’s built-in retry policy.
- Simplify the connector’s event payload construction and cursor handling logic.

Documentation:
- Document the new connector behavior, throttling, and paging semantics in the changelog.

Tests:
- Update and extend tests to cover page-size capping, batched streaming behavior, unified throttling, and the revised cursor semantics.

Chores:
- Bump the Flare.io integration version to 0.5.0 in the manifest.

### DIFF
--- a/Flare/CHANGELOG.md
+++ b/Flare/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-07-20 - 0.5.0
+
+### Changed
+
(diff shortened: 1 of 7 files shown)


TOUFIKIzakarya (approved):
LGFM

PR #2917, opened 2026-07-22
Sekoia.io: handle null value

squioc:
Fix the way the trigger handles null values in some parameters of an alert.

## Summary by Sourcery

Handle null or absent alert attributes safely in Sekoia.io triggers and document the fix.

Bug Fixes:
- Ensure alert triggers correctly process events when message.attributes is null or missing.
- Prevent failures when alert status, rule, verdict, custom_status, urgency, entity, alert_type, or rule fields are null in alert payloads.

Enhancements:
- Improve alert trigger robustness by consistently defaulting nullable nested fields to empty dictionaries before access.

Build:
- Bump Sekoia.io module version to 2.74.5.

Documentation:
- Add changelog entry describing the fix for handling absent or null values in triggers.

Tests:
- Extend alert trigger tests to cover alerts and comments with null attributes and nested fields.

### DIFF
--- a/Sekoia.io/CHANGELOG.md
+++ b/Sekoia.io/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-07-22 - 2.74.5
+
+### Fixed
+
+- Fix the way to handle absent or null values in the triggers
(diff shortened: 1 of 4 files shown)


PierrickV (commented):
Shouldn't this be a pydantic model ? 🙏

PierrickV (approved):

PR #2926, opened 2026-07-27
feat(EntraID): add reset checkpoint method

TOUFIKIzakarya:
Linked issues and PRs :
1 - https://github.com/SekoiaLab/platform/issues/89125
2 - https://github.com/orgs/SekoiaLab/projects/90/views/32

## Summary by Sourcery

Add support for resetting the Entra ID asset connector checkpoint and expose field mappings for user assets, with corresponding tests and version bump.

New Features:
- Introduce a reset_checkpoint method on the Entra ID asset connector to clear stored user asset checkpoints.
- Expose a get_mapped_fields API returning the OCSF field mappings for Entra ID user assets.

Build:
- Bump the Microsoft Entra ID integration manifest version from 2.10.23 to 2.10.24.

Documentation:
- Update the Microsoft Entra ID changelog with a new 2.10.24 release entry describing the added methods.

Tests:
- Add async tests covering reset_checkpoint behavior with and without an existing checkpoint.
- Add a test verifying that get_mapped_fields returns a non-empty and correctly structured mapping dictionary.


MicrosoftEntraID/azure_ad/asset_connector/user_assets.py
```diff
@@ -98,6 +98,30 @@ async def update_checkpoint(self) -> None:
...
+        self._latest_time = None
+
+    def get_mapped_fields(self) -> dict[str, str]:
+        return {
+            "id": "user.uid",
+            "displayName": "user.full_name",
+            "mail": "user.email_addr",
+            "userPrincipalName": "user.name",
+            "accountEnabled": "enrichments.account.data.is_enabled",
+            "createdDateTime": "time",
```
StephKll3c: `createdDateTime` maps to `"time"` here, but `created_date_time` isn't used anywhere in `map_fields` and no `time=` is set on the `UserOCSFModel`, so it isn't actually mapped. Same for `employeeId`/`employeeType` mapping to `enrichments.employment.data`: in `map_fields` these only gate the `if`, and `employment_data = UserDataObject()` is empty, so they're never stored. Either this dict is inaccurate or `map_fields` is missing those mappings. Worth aligning, since checkpoint reset keys off this.

MicrosoftEntraID/tests/asset_connector/test_users_assets.py
```diff
@@ -997,3 +997,54 @@ def test_map_fields_enrichment_without_optional_fields(test_entra_id_asset_conne
...
+
+    # Act — should not raise
+    await test_entra_id_asset_connector.reset_checkpoint()
+
+    # Assert
+    assert test_entra_id_asset_connector.most_recent_date_seen is None
+    assert test_entra_id_asset_connector._latest_time is None
+
+
+def test_get_mapped_fields(test_entra_id_asset_connector):
```
StephKll3c, line 1032: This only checks the values are non-empty strings, so it wouldn't catch a drift between `get_mapped_fields` and the actual `map_fields` output. Could we pin the exact expected mapping (snapshot) here?

mchupeau-sk (approved):
Seems good to me ! LGFM

PR #2940, opened 2026-07-29
fix(sentinelone): duplication of events

TOUFIKIzakarya:
Main issue : [issue](https://github.com/SekoiaLab/platform/issues/89962)

## Summary by Sourcery

Prevent duplicate SentinelOne events by persisting the events cache after each paginated API page instead of only on connector shutdown.

Bug Fixes:
- Avoid event duplication after unexpected pod restarts by saving the session events cache incrementally during event pulling.

Enhancements:
- Improve resilience of the SentinelOne connector by ensuring the events cache survives unclean shutdowns.

Build:
- Bump SentinelOne integration manifest version from 1.23.0 to 1.23.1.

Documentation:
- Update SentinelOne changelog with version 1.23.1 and the corresponding fix description.

### DIFF
--- a/SentinelOne/CHANGELOG.md
+++ b/SentinelOne/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 2026-07-29 - 1.23.1
+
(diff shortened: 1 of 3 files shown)


mchupeau-sk (approved):
LGFM

PR #2943, opened 2026-07-30
fix(microsoft_defender): fix duplication

TOUFIKIzakarya:
main issue : [issue](https://github.com/SekoiaLab/platform/issues/90667)

## Summary by Sourcery

Preserve Microsoft Defender asset connector checkpoints using the raw lastSeen timestamp string to avoid asset duplication caused by precision loss.

Bug Fixes:
- Prevent duplicated assets by storing the Defender lastSeen timestamp checkpoint as the original ISO string instead of converting to datetime and truncating precision.

Enhancements:
- Track the most recent lastSeen value in get_assets using string-based comparison via parsed timestamps while retaining the exact original value.

Build:
- Bump the Microsoft Defender connector manifest version from 1.3.3 to 1.3.4.

Documentation:
- Update the changelog with a new 1.3.4 entry describing the timestamp checkpoint fix.

Tests:
- Extend device asset connector tests to cover raw lastSeen checkpoint storage, precision preservation, and end-to-end checkpoint usage across get_assets and update_checkpoint.

### DIFF
--- a/MicrosoftDefender/CHANGELOG.md
+++ b/MicrosoftDefender/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## 2026-07-30 - 1.3.4
+
+### Fixed
+
+- Store the raw `lastSeen` string from the API directly as checkpoint instead of converting to `datetime` and back.
+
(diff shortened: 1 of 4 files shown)


mchupeau-sk (dismissed):
Some errors on test

mchupeau-sk (approved):

PR #3027, opened 2026-08-25
fix(asset_connector): delete just one field

TOUFIKIzakarya:
## Summary by Sourcery

Remove the Microsoft Entra ID user creation timestamp mapping from asset fields.

Bug Fixes:
- Stop mapping the Microsoft Entra ID user creation timestamp to the asset `time` field, ensuring only the intended field is removed.

Tests:
- Update asset connector mapping expectations to reflect removal of the creation timestamp mapping.

Chores:
- Bump the Microsoft Entra ID connector version to 2.10.27 and document the change in the changelog.


mchupeau-sk (approved):
LGFM

PR #3049, opened 2026-08-27
fix sekoia module

mchupeau-sk:
## Summary by Sourcery

Prevent SSL recursion errors by limiting early gevent patching to trigger executions.

Bug Fixes:
- Fix HTTPS trigger executions by applying gevent monkey-patching before SSL-related modules are loaded.

Chores:
- Bump the Sekoia.io module version to 2.76.2.

### DIFF
--- a/Sekoia.io/main.py
+++ b/Sekoia.io/main.py
@@ -1,7 +1,9 @@
 # flake8: noqa: E402
-from sekoiaio.utils import should_patch
+import sys
 
-if should_patch():
+# Patch before any import pulls in ssl (via requests/urllib3), otherwise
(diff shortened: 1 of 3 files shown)


Sekoia.io/main.py
```diff
@@ -1,7 +1,9 @@
 # flake8: noqa: E402
-from sekoiaio.utils import should_patch
```
TOUFIKIzakarya, line 2: I think there's also the `base.py`. And if should_patch is not needed anymore we can delete it

TOUFIKIzakarya (approved):

PR #3054, opened 2026-08-31
Fix: Holm Devices And Vulnerabilities (1902)

vg-svitla:
Relates to [1902](https://github.com/SekoiaLab/integration/issues/1902)

## Summary by Sourcery

Make Holm asset and vulnerability collection resilient to nullable API responses while preserving accurate asset data.

Bug Fixes:
- Prevent device and network asset collection from failing when Holm returns explicit null values for optional fields.
- Preserve null and zero risk scores correctly when mapping assets.
- Parse device email recipients as objects containing email and username.
- Accept the published Holm API shapes for nullable tags, vulnerability data, and asset fields.

Enhancements:
- Centralize nullable Holm API model handling while continuing to reject missing required identifiers.

Tests:
- Add coverage for nullable API responses, email object parsing, minimal tags, identifier validation, and risk-score mapping.

Chores:
- Bump the Holm Security integration version to 1.3.1 and update the changelog.

### DIFF
--- a/HolmSecurity/CHANGELOG.md
+++ b/HolmSecurity/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [1.3.1] - 2026-08-28
+## [1.3.1] - 2026-09-03
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


TOUFIKIzakarya (commented):
- With this fix, if the API returns `None` for a field like `risk_score` (int = 0), the key gets removed and falls back to 0.
Could you confirm if setting `risk_score`= 0 for None payloads is the desired behavior, or should we preserve None (e.g. using int | None = None) to differentiate a score of zero from an absent score?
- Can you also check the [doc](https://se-api.holmsecurity.com/docs/v2/#tag/devices/operation/devices_list) for required fields and optional fields

TOUFIKIzakarya (approved):
LGFM

PR #3126, opened 2026-09-18
Fix/aws memory optimization

squioc:
- use generators to improve memory usage in connectors
- use [isjon](https://github.com/ICRAR/ijson/) to read large JSON document in the S3 records connector

## Summary by Sourcery

Optimize AWS S3 connectors for memory-efficient processing of large objects.

Enhancements:
- Reduce memory usage when processing large S3 objects by streaming log and event records through connector pipelines.
- Parse S3 JSON records incrementally to avoid loading complete documents into memory.

Build:
- Add ijson as an AWS connector dependency and update the AWS module to version 1.38.2.


TOUFIKIzakarya (approved):
Thanks for this optimization.
LGFM

